PR #18, opened 2022-12-22
Moved score, accuracy + misses text info text below the hp bar for downscroll

NebulaZone1:
Having the info text below the hp bar when downscroll is on to make it a bit more easier to read than having it on top of the healthbar


NebulaZone1:
If you don't want this, feel free to close it

YoshiCrafter29:
Prefer not to merge, might add an option to put the score text above the strums tho

NebulaZone1:
> Prefer not to merge, might add an option to put the score text above the strums tho

Alright 👍

PR #23, opened 2022-12-24
Losslessly compress the png's and ogg's

Randomxdp:
hi, so i just wanted to compress the ogg's and png's to reduce the size of the engine, the tools i used were oxipng v8 and optivorbis (latest), just in case you wanna use those to optimize some png or ogg added to the engine after this pr, i checked the png files quickly with pngcheck to see if they didnt get corrupted or smth, but still try putting the files from this pr into a yoshi engine build to see if they dont cause any issues before merging, because i optimized with my phone via termux and compiling optipng and oxipng with cargo (call me insane lol), some sprites (gf christmas for example) arent max optimized because of their insane ram usage, but i tried optimizing them a bit as well


Randomxdp:
i can remove the compressed ogg files if you want and only leave the pngs

FoxelTheFennic:
im not able to test this build right now, but everything looks good to me! 👍👍

@YoshiCrafter29 whatcha think abt possibly automating this on build in the future? probably would be not that useful but seeing as how much space fnf already takes up i'd be nice to have

YoshiCrafter29:
Merging this since it can actually boost up the engine.
I'll also prob write a macro to automatically do this perhaps, might check
Thanks!

Randomxdp:
no problem, just wanted to help :)

Randomxdp:
as for the macro i think it should be a precommit hook

PR #33, opened 2022-12-27
Minor Spelling Mistake

BombasticTom:
I assume you meant `FlxG.sound.soundTrayEnabled = true;` instead of `FlxG.sound.soundTrayEnabled = ();`. Correct me if I'm wrong, that was just my assumption.


BombasticTom:
Sorry, I didn't notice that the issue was resolved :P

PR #44, opened 2023-01-07
Fix game crashing on mac

superplayhero2:
build number crashes the game when it's launched
and current hxCodec version doesn't support mac // yoshicrafter29 fixed the hxCodec thing
so i made this pull request to fix that


YoshiCrafter29:
Fixed by NeeEoo

PR #61, opened 2023-01-18
Separeted and character dependent voices

TomyGamy:
Allow the voice to be change depending on the played character, also separated voices to make the opponent not stop singing when the player miss (examples on the test song)


lunarcleint:
you can add this in hscript bro...,,,..

TomyGamy:
On the Charting state? (WIP)
This is very useful in this tool

biomseed:
this is how the base game update will handle it actually

lunarcleint:
you can add it in charting with hscript

PR #65, opened 2023-01-23
Very minor change but...

SrtHero278:
Can you merge this so scripts in `assets/data/charts/` apply in every song?
It's annoying since I have some global scripts that I wanna to be able to use in mods.
Not to mention that with how it is currently, you have to put a script in every mod if you want to use a fanmade editor.


BushTrain460615:
what about people who want to use charts on specific songs?

SrtHero278:
> what about people who want to use charts on specific songs?

This is for PlayState global scripts. Song specific scripts are in `songs/[SONG-NAME]/scripts`.

BushTrain460615:
ok my mistake

lunarcleint:
Already added in

PR #66, opened 2023-01-24
Made update file shell executable for linux

TumoPlayz:
you may not see anything but now people can run "./update.sh" and it will run


TumoPlayz:
this aged like milk

TumoPlayz:
re opened because read me is misleading

TumoPlayz:
the folder for common commands is /bin/bash not /bin/sh

lunarcleint:
alreadly added

TumoPlayz:
OH

PR #68, opened 2023-01-27
put substate scripts in their own folder

swordcube:
it's annoying to have states and substates in the same exact folder


swordcube:
literally a 3 letter fix

RafGamign:
g 🌮

Stilic:
bad idea since it can break existing scripts
people will never guess the problem bruh

swordcube:
> bad idea since it can break existing scripts people will never guess the problem it causes at the end bruh

i could also make the change backwards compatible

lunarcleint:
substates and states are un-separated in source so why should they be here?

PR #90, opened 2023-02-26
minor Note.hx optimization

Burgerballs:
`FlxSprite` has unnecessary calculations for velocity and other stuff if the variable `moves` isn't set to `false`, if the mod creator wants to make use of velocity again they can set it back to true, so what I simply did was `moves = false;`.


Rom4chek (approved):

SirAxolot:
couldnt you just do `moves = false;` on every note when it's created instead? so theres no need for this to be hardcoded

lunarcleint:
Alreadly added

PR #117, opened 2023-04-23
Fix Week 7 Cutscenes

SwickTheGreat:
Fixes a bug where the week 7 cutscenes wouldn't play and it would soft-lock the game.


NeeEoo:
This wont work, since the sprites need a FunkinSprite, if you simply update the code to be compatible with the new source, then you dont need to change have the import

SwickTheGreat:
Alright, will fix!

SwickTheGreat:
Cloasing this PR, since it is fixed in the latest commit.

PR #124, opened 2023-05-15
Fixes for Linux

ItzFireable:
Latest commits broke the Linux builds, so here's a pull request to fix said issues.

- Fixes a typo in one of the HiddenProcess calls.
- Throws an exception if CPU is not found on Linux, would not compile due to it before.
- Change Compiler.hx to target current build target on all platforms instead of Windows only.
- Added the Consolas font into assets/fonts, fixing the font not working on Linux and most likely Mac too.

(99% sure that the usage is allowed, but if its not, I'll get rid of it. Read the [Microsoft font FAQ](https://learn.microsoft.com/en-us/typography/fonts/font-faq) to make sure it is allowed.)


ItzFireable:
Alongside this PR, I recommend accepting #123 so action builds work again (the errors in the Linux action there is due to these issues).

PR #129, opened 2023-05-24
Strum Extensibility

Dirk-Dinkum:
This PR allows you to add more than 4 strums to a strumline via scripting, while also changing its keybinds.

For example, this script below allows you to add a fifth strum (rotated up strum) controlled by the spacebar on the player's side, and each note has a 10% chance of being on that new strum.
image
```haxe
function onPostGenerateStrums() {
	for(i in 0...2)
		player.members[i].x -= Note.swagWidth * 0.5;
	for(i in 2...4)
		player.members[i].x += Note.swagWidth * 0.5;

	var str = player.createStrum(4, "up");
	str.x -= 2.5 * Note.swagWidth;
	str.noteAngle = 0;
	str.angle = 45;
	str.getPressed = function() {
		return FlxG.keys.pressed.SPACE;
	};
	str.getJustPressed = function() {
		return FlxG.keys.justPressed.SPACE;
	};
	str.getJustReleased = function() {
		return FlxG.keys.justReleased.SPACE;
	};
}

var lastChoice = false;

function onPostNoteCreation(event) {
	if (event.note.strumLine != player) return;

	if (event.note.isSustainNote) {
		if (lastChoice)
			event.note.noteData = 4;
	} else if (lastChoice = FlxG.random.bool(10)) {
		event.note.noteData = 4;
	}
}
```

### DIFF
--- a/source/funkin/backend/scripting/events/StrumCreationEvent.hx
+++ b/source/funkin/backend/scripting/events/StrumCreationEvent.hx
@@ -22,6 +22,11 @@ final class StrumCreationEvent extends CancellableEvent {
 	public var strumID:Int;
 
 	/**
+	 * Animation prefix (`left` = `arrowLEFT`, `left press`, `left confirm`).
+	 */
+	public var animPrefix:String;
+
+	/**
 	 * Sprite path, in case you only want to change the sprite.
 	 */
 	public var sprite:String = "game/notes/default";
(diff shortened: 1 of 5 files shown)


WizardMantis441:
OOUGHHH DIRK THIS IS PEAK

PR #133, opened 2023-05-31
Re-Export Sprites With Padding

SwickTheGreat:
Re-exported `gf-car`, `gf-christmas`, `gf-tankmen`, and `pico-speakers` with 1 pixel border padding and 3 pixel symbol padding.


SwickTheGreat:
Do not merge, there are offset issues I need to fix!!

PR #159, opened 2023-07-01
Fix Score being increased while "showRating" is disabled and on sustains

BeastlyGabi:
beep


Sword352 (approved):
I have a farm and a cow called suzie

lunarcleint:
W

PR #178, opened 2023-07-05
Make offset points get put in the pool when FunkinSprite is destroyed

MemeHovy:
thanks to @Sword352 for the assist


MemeHovy:
> slay

ikr

Sword352 (approved):
SLAAAAY

MemeHovy:
> SLAAAAY

🥰

NeeEoo:
Add a null check on the animOffsets in destroy, just to prevent a crash if you destroy it twice

MemeHovy:
> Add a null check on the animOffsets in destroy, just to prevent a crash if you destroy it twice

alright 👍🏻

PR #184, opened 2023-07-16
Note types in charter

bctix:
makes it possible to add, remove, place, and change a notes type in the chart

https://github.com/FNF-CNE-Devs/CodenameEngine/assets/51544115/69c84a92-4a47-4ba9-83fe-76e1400085a9

### DIFF
--- a/source/funkin/editors/charter/Charter.hx
+++ b/source/funkin/editors/charter/Charter.hx
@@ -227,6 +227,16 @@ class Charter extends UIState {
 					},
 					null,
 					{
+						label: "Switch to previous note",
+						keybind: [ONE],
+						onSelect: _note_prevnote
+					},
+					{
+						label: "Switch to next note",
+						keybind: [TWO],
+						onSelect: _note_nextnote
+					},
+					{
 						label: "Edit note types",
 						keybind: [],
 						onSelect: _edit_notetype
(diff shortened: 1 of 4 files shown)


MemeHovy (approved):
this cool af

source/funkin/editors/charter/CharterNote.hx
```diff
@@ -131,5 +141,6 @@ class CharterNote extends UISprite {
 	public override function draw() {
 		drawMembers();
 		drawSuper();
+		numberLabel.draw();
```
NeeEoo, line 144: Id recommend doing a if(type != 0)

Because it would increase performance

source/funkin/editors/charter/Charter.hx
```diff
@@ -225,12 +227,9 @@ class Charter extends UIState {
...
 					null,
 					{
-						label: "(0) Default Note",
-						keybind: [ZERO]
-					},
-					{
-						label: "(1) Hurt Note",
-						keybind: [ONE]
+						label: "Edit note types",
+						keybind: [],
```
NeeEoo, line 231: You should make it so it says

[0] default
[1] hurt
[2] death
...

And like so when you press a number key it switches.

Or like make it so

[0] default
[1] go to next note type
[2] go to previous note type

NeeEoo (commented):
Looks good! But i have some suggestions

NeeEoo:
Make it say notetype, also you forgot the 0 to switch to the default note

NeeEoo:
Oh and maybe make sure that you cant delete the default note type

lunarcleint:
I say the drop down is cool but I still think the old way where it was in the top menu notes tab looks 100x better

PR #191, opened 2023-07-18
Camera Speed Within XML.

teddytony:


teddytony:
> Isn't it a little useless since you can change `FlxG.camera.followLerp` in the hx stage file?

some prob don't know how `FlxG.camera.followLerp` works, but I see where your coming from

NeeEoo:
i dont really understand why you are changing the speeds of all of the stages.

Also personally i would use a setter too

Also dont set it after the create, set it at line 666 instead, also use 0.04 instead of 0.05

teddytony:
> i dont really understand why you are changing the speeds of all of the stages.
> 
> Also personally i would use a setter too
> 
> Also dont set it after the create, set it at line 666 instead, also use 0.04 instead of 0.05

I really don't know why I did change the stages when their going to have an default value anyway.

PR #193, opened 2023-07-20
get mod states working

CharlesCatYT:
image

image

probably reverts the script pack thing, mod states should work again when merged lol
yea this isnt a good explanation since this is pretty much my first actual pr but should do ig
(its drafted for a reason cuz i dont know if theres any problems to solve here i think)


lunarcleint:
Looks good thanks

PR #198, opened 2023-08-21
Add Quantization to the chart editor

Burgerballs:
image

adds note quantization to the chart editor

Press Z or X to decrease and increase the beat snap respectively

It also resizes the grid so it's easier to see where you should place the notes :)


lunarcleint:
Sorry alr added in update, thanks for your time tho :((

PR #229, opened 2023-12-22
Updated the video backend to support the latest `hxvlc` release.

MAJigsaw77:


MemeHovy:
see https://github.com/FNF-CNE-Devs/CodenameEngine/pull/227

quoted from @NexIsDumb 
> the video library is receiving many updates though @NeeEoo is stuck with the older version cause he wants to use that for now
i already asked him and he told me to use the commit he was using on that branch instead of making changes about it in the source

PR #241, opened 2024-02-07
Add array option type

bctix:
adds an array option type :)

https://github.com/FNF-CNE-Devs/CodenameEngine/assets/51544115/5454da64-c460-40d6-afba-b3ea8815eb68

### DIFF
--- a/assets/data/config/options.xml
+++ b/assets/data/config/options.xml
@@ -4,9 +4,9 @@
 	<checkbox id="checkboxExample" name="Checkbox example" />
 	<number id="numberExample" name="Number example" min="0" max="10" change="1"/>
 	<choice id="choiceExample" name="Choice Example">
-		<value name="Disabled" />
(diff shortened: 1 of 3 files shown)


lunarcleint:
This is a REALLY nice pull request, we've been meaning to add this for awhile. Just can you explain why you chose to use the inner data of the xml attributes instead of using the original data structure? It clashes with the rest of the xml

bctix:
> This is a REALLY nice pull request, we've been meaning to add this for awhile. Just can you explain why you chose to use the inner data of the xml attributes instead of using the original data structure? It clashes with the rest of the xml

Ne_Eo asked me to structure the xml like that.

lunarcleint:
Me and Ne_Eo talked about it and if you change it back we will merge you pull request :D, sorry for the time loss 😭

lunarcleint:
thank you for cool pull :D

PR #243, opened 2024-02-15
Fix repeating number option

bctix:
fixes a problem that I would come across when using the number option where if using decimal numbers, there's a chance ( i don't know what causes it exactly ) for the number to be repeated

before:

image

after:

image

### DIFF
--- a/source/funkin/options/type/NumOption.hx
+++ b/source/funkin/options/type/NumOption.hx
@@ -49,6 +49,7 @@ class NumOption extends OptionType {
 	{
 		if(currentSelection <= min && change == -1 || currentSelection >= max && change == 1 ) return;
 		currentSelection += change*changeVal;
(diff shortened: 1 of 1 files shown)


Sword352 (approved):
math can be silly somtimes but this does the trick

NexIsDumb:
holy shit good fucking point
i'll get my pc in a few minutes, edit some stuff for optimizations and merge it, thanks!

NexIsDumb:
tyy!

Sword352 (commented):
hold on sir you're getting the decimals before changing the value will it work 🤨

NexIsDumb:
> hold on sir you're getting the decimals before changing the value will it work 🤨 

yeah cause its rounding the math inside the first argument and getting the decimals from the value inside the xml

PR #245, opened 2024-02-18
Simple small charter fixes

EstoyAburridowGH:
1. Small fix for a slightly annoying error about the charter trying to load the voices even if you marked them as not necessary
image
image

2. The charter did not detect icons that did not have the same name as the character
Before:
image
After:
image


NexIsDumb:
right! tysmm

PR #249, opened 2024-03-02
OpenFL version compatibility fix

Jamextreme140:
Uses var alias "__source" on 9.2.2 or "__audioSource" on latest, in case of using latest libraries.

PR #250, opened 2024-03-03
OpenFL version compatibility fix (reopen)

Jamextreme140:
(reopen of #249 with a new solution)

Uses (compiles) "__source" or "__audioSource" variable depending of the OpenFL version, in order to not break FlxSound.

### DIFF
--- a/source/openfl/media/SoundChannel.hx
+++ b/source/openfl/media/SoundChannel.hx
@@ -63,7 +63,11 @@ import lime.media.AudioSource;
 	@:noCompletion private var __isValid:Bool;
 	@:noCompletion private var __soundTransform:SoundTransform;
 	#if lime
+		#if (openfl < "9.3.2")
 	@:noCompletion private var __source:AudioSource;
+		#else
+	@:noCompletion private var __audioSource:AudioSource;
+		#end
 	#end
 
 	#if openfljs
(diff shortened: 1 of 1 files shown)


NexIsDumb:
thats not the getter and setter is, we meant the one that haxe gives you as possibility
want me to make a commit to your pr to help ya?

Jamextreme140:
> thats not the getter and setter is, we meant the one that haxe gives you as possibility want me to make a commit to your pr to help ya?

Thanks, but now I think I got it. Check the last commit.

NexIsDumb:
oh ye you did it :D
though i think itd be better if you added __audioSource only for the new openfl, so like do the compiler check and its good :D

Jamextreme140:
I don't know what did you meant with that.

The reason of this PR is because the overriden `openfl.media.SoundChannel` class in source has conflict with the installed library on haxelib (in case you decide to upgrade it), specifically I got this error: `FlxSound.hx:[...] openfl.media.SoundChannel has no field "__audioSource"`

NexIsDumb:
hmm alr then lets leave it like that, you tested everything right?

Jamextreme140:
Yeah, compiles and game running fine

NexIsDumb:
great then, just tested aswell, looks great :D
tysm for ur time on the enginee

PR #257, opened 2024-03-17
FunkinParentDisabler: Prevents camera movement, shaking, and update flash

EstoyAburridowGH:
Well, as the title mentions, this is a simple change to the FunkinParentDisabler class, which makes it so that when you are, for example, in the pause menu, the camera, in addition to stopping following, stops shaking and stops updating the flash

### DIFF
--- a/source/funkin/backend/utils/FunkinParentDisabler.hx
+++ b/source/funkin/backend/utils/FunkinParentDisabler.hx
@@ -32,8 +32,8 @@ class FunkinParentDisabler extends FlxBasic {
 			FlxTimer.globalManager._timers = [];
 
 			// cameras
-			__cameras = [for(c in FlxG.cameras.list) if (c.followEnabled) c];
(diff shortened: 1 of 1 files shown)


CharlesCatYT (approved):
nex gotta see this i want it now

NexIsDumb:
ooo
the idea sounds good :D
tmr we should also merge the dev repo of cne so i'll surely test this pr out after that and merge it if everything is good :D

NexIsDumb:
OKAY I can finally merge this
im sorry it took me almost 2 weeks😭
tyy :D

NexIsDumb:
uh oh it seems this pr is causing issues in the pause menu
for example if you try to resize the windows in the pause menu the cameras wont resize, we might revert this and make each thing toggable instead prob
image

EstoyAburridowGH:
:sob:

PR #281, opened 2024-04-13
FunkinParentDisabler: Fix camera not updating its size

EstoyAburridowGH:
In my previous pull request I proposed a way to prevent the camera from continuing to update but it also brought another bug preventing the camera from updating its size when changing the window size
Now in this pull request I have another simple change that should fix it (I hope this doesn't give us another bug lol)
### Before:
image
### After:
image


NexIsDumb:
oh i stg I was about to start fixing it aswell literally in 1h LOL

NexIsDumb:
Imma start including this in the newer update tyy :D

NexIsDumb:
ty!!

Issue #284, opened 2024-04-17
Bug Report: custom alphabet doesn't reload


what-is-a-git:
[weed]: <> (MAKE SURE THAT THE ISSUE IS AS COMPLETE AS POSSIBLE. ISSUES THAT HAVE BARELY ANY INFORMATION OR IS ONLY MADE FOR TROLLING WILL BE INSTANTLY CLOSED.)
[weed]: <> (MAKE SURE TO ADD YOUR TARGET OR IT'LL BE IGNORED!)

### Which target are you using?
[weed]: <> (CHANGE [ ] TO [x] IN FRONT OF YOUR TARGET!)

[x] Windows
[ ] Linux
[ ] Mac

### Explain your issue

basically when you use a custom data/alphabet.xml it doesn't get reloaded when loading into other mods, causing some strange behavior if you use differently named animations

![image](https://github.com/FNF-CNE-Devs/CodenameEngine/assets/138817584/f4da314b-0419-4c01-94a5-f08e706e4d0a)

also fun fact: you can crash the game if you rename the custom alphabet from alphabet.xml to something else and load into another mod

![image](https://github.com/FNF-CNE-Devs/CodenameEngine/assets/138817584/8d7e9225-fc17-42b7-b694-ffa7b58d07a7)

usb-port-2:
"you can crash the game if you rename the custom alphabet from alphabet.xml to something else and load into another mod"
no shit dumbass it needs that to run the game

if i delete system32 it isn't window's fault my pc doesn't work, it's mine because i was being a dumbass

moxie-coder:
> "you can crash the game if you rename the custom alphabet from alphabet.xml to something else and load into another mod" no shit dumbass it needs that to run the game
> 
> if i delete system32 it isn't window's fault my pc doesn't work, it's mine because i was being a dumbass

Ikr

usb-port-2:
"causing some strange behavior if you use differently named animations"
No fucking way, using something incorrectly does something wrong? I had no idea! I will tell the codename engine devs to cancel the entire thing and delete the repository.

This isn't an issue with the engine, dude.

what-is-a-git:
it seems my original post may have not been that clear on what the issue i'm talking about really is so i'm making this comment to clarify

the issue was never about data/alphabet.xml not working, it works as you'd expect, allowing you to add, change, and remove different animations in alphabet text.

[...]

PR #288, opened 2024-04-26
fix alphabet stuff from #284

what-is-a-git:
this just fixes the whole issue in 2 lines of code

i'm not even kidding

it wasn't reloading cuz uhh Paths.xml doesn't have cool mods folder stuff in it, so if u just like add that and check for it it just works

:D 👍


NexIsDumb:
oh shoot ty!!

PR #293, opened 2024-05-10
Don't create the vocals' waveform if they weren't loaded

Stilic:
Fixes that one infamous crash on the charter that forces people to use older versions of CNE.


NexIsDumb:
yeah thanks for making me notice
i kinda noticed how the needsvoices part was kinda rushed
i wanted to fix it in dev and at this point imma use this pr as a starting point
so i might merge this before releasing the update
tysm!!

NexIsDumb:
okay so I noticed that @Frakits already fixed this, still ty!!

PR #306, opened 2024-06-01
Safe Mode implementation

Jamextreme140:
It disables the usage of classes that can perform malicious operations like Sys.command() and cpp.Lib.load().

**Reflect** and **Type** are disallowed since it can access to blacklisted classes (Reflect.callMethod() and Type.resolveClass())

Example: A song script tries to execute this following song script (function postCreate()):
`trace(Reflect.getProperty(PlayState, "difficulty"))`
`Sys.command("echo Howdy")`

Captura de pantalla 2024-06-01 124626
After disabled "Safe mode" it works normally again

_Safe Mode mark when it's enabled_
Captura de pantalla 2024-06-01 124528

### DIFF
--- a/source/funkin/backend/scripting/HScript.hx
+++ b/source/funkin/backend/scripting/HScript.hx
@@ -38,9 +38,6 @@ class HScript extends Script {
 		interp.importFailedCallback = importFailedCallback;
 		interp.staticVariables = Script.staticVariables;
 		interp.allowStaticVariables = interp.allowPublicVariables = true;
-		#if SAFE_MODE
-		interp.importBlocklist = Script.disallowedimports;
-		#end
 
 		interp.variables.set("trace", Reflect.makeVarArgs((args) -> {
 			var v:String = Std.string(args.shift());
--- a/source/funkin/backend/scripting/HScript.hx
+++ b/source/funkin/backend/scripting/HScript.hx
@@ -38,6 +38,9 @@ class HScript extends Script {
 		interp.importFailedCallback = importFailedCallback;
 		interp.staticVariables = Script.staticVariables;
 		interp.allowStaticVariables = interp.allowPublicVariables = true;
+		#if SAFE_MODE
+		interp.importBlocklist = Script.disallowedimports;
+		#end
 
 		interp.variables.set("trace", Reflect.makeVarArgs((args) -> {
 			var v:String = Std.string(args.shift());
--- a/source/funkin/backend/scripting/HScript.hx
+++ b/source/funkin/backend/scripting/HScript.hx
@@ -38,6 +38,10 @@ class HScript extends Script {
 		interp.importFailedCallback = importFailedCallback;
 		interp.staticVariables = Script.staticVariables;
 		interp.allowStaticVariables = interp.allowPublicVariables = true;
+		if(Options.safemode)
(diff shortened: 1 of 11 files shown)


NeeEoo:
while(Script.importBlocklist.length > 0) Script.importBlocklist.pop()

NexIsDumb:
and i'm also not sure about this either, well also first of all cause its easily avoidable as neeo made you see but also cause it would create MANY limitations

Jamextreme140:
Wait. I think I have a better idea for Safe Mode

NeeEoo:
We came up with a idea where like  its gonna be like a toggle thats by default off

Jamextreme140:
Yeah, that's exactly what I was thinking about. Safe Mode as an Option instead of a separated build

usb-port-2:
This _**SUCKS**_ so _**GODDAMN MUCH**_.

I should be able to _**DESTROY**_ your **_PC_** via my **_SILLY_** and **_FUNNY_** rhythm game song.

Jamextreme140:
> We came up with a idea where like its gonna be like a toggle thats by default off

check you the latest commit

source/funkin/backend/scripting/Script.hx
```diff
@@ -194,6 +205,8 @@ class Script extends FlxBasic implements IFlxDestroyable {
 		this.path = path;
 		onCreate(path);
 		for(k=>e in getDefaultVariables(this)) {
+			if(getBlacklistedImports().contains(k))
```
NeeEoo, line 208: Id recommend calling this once just outside of the loop

source/funkin/backend/scripting/HScript.hx
```diff
@@ -38,6 +38,10 @@ class HScript extends Script {
 		interp.importFailedCallback = importFailedCallback;
 		interp.staticVariables = Script.staticVariables;
 		interp.allowStaticVariables = interp.allowPublicVariables = true;
+		if(Options.safemode)
+		{
+			interp.importBlocklist = Script.getBlacklistedImports();
```
NeeEoo, line 43: Btw

`__script__.interp.importBlocklist` is accessable in a script

source/funkin/backend/scripting/Script.hx
Jamextreme140: Do you mean creating a separate array to use it or create a map from getDefaultVariables and modify it?

source/funkin/backend/scripting/HScript.hx
Jamextreme140: well, there is the problem. I see it's not possible to deny the access to "importBlocklist" without breaking the entire engine.

source/funkin/backend/scripting/Script.hx
NeeEoo: Nah just like

var blockedImports = Script.getDefaultImports()
for k >= in aaaaaaa:
Im tooo tired rn but you get me

NeeEoo:
We could probably just do @:unreflective on hscript-improved to turn off being able to edit it

Jamextreme140:
> We could probably just do @:unreflective on hscript-improved to turn off being able to edit it

Hmm, it can work

PR #307, opened 2024-06-01
FPS Counter offset fix for Options/Credits menu

maplesyruppppppp:
pretty self explanatory
will also work for when the pathBG height is extended
image


NexIsDumb:
oo wait i love this

NexIsDumb:
ty!!

PR #311, opened 2024-06-04
Mall fix (why was it like that)

Kolziak:
just made the mall look like in vslice


NexIsDumb:
oh? lemme check rq

NexIsDumb:
oh thats true, although lemme just

NexIsDumb:
ty!!

PR #313, opened 2024-06-07
removed unused cpu thingy line of code

MarioGuyTakenAgain:


NexIsDumb:
ermmm, no its actually very used
take a look at line 68 and 69 of that same file just for example

NexIsDumb:
ermmm, no its actually very used
take a look at line 68 and 69 of that same file just for example

PR #323, opened 2024-06-24
no more new button in the charter

maplesyruppppppp:
i think it should be removed cause it does nothing and we already have a way of making new song files in the engine itself

before
image

after
image


NexIsDumb:
oh wait thats actually true😭
@lunarcleint

NexIsDumb:
oh wait nvm we had a better idea that we implemented in dev, we'll merge it once the update is out
ty anywaysss

maplesyruppppppp:
> oh wait nvm we had a better idea that we implemented in dev, we'll merge it once the update is out ty anywaysss

o alr lmao

PR #337, opened 2024-07-06
Fix Null object Reference

TheoDevelops:
When you pause in a specific time in the countdown (when the countdown is finishing) it crashs.

I just added a null check for the song to fix it.

### DIFF
--- a/source/funkin/game/PlayState.hx
+++ b/source/funkin/game/PlayState.hx
@@ -1260,7 +1260,7 @@ class PlayState extends MusicBeatState
 				if (Conductor.songPosition >= 0)
 					startSong();
 			}
-		} else {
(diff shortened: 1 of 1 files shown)


source/funkin/game/PlayState.hx
```diff
@@ -1260,7 +1260,7 @@ class PlayState extends MusicBeatState
 				if (Conductor.songPosition >= 0)
 					startSong();
 			}
-		} else {
+		} else if (!startingSong && FlxG.sound.music != null) {
```
NeeEoo, line 1263: The !startingSong isn't needed due to the if condition above

dotaxel: letme fix it

dotaxel: done

PR #359, opened 2024-07-28
Inactive volume

bctix:
Adds the feature from osu!lazer where when the window is inactive, the volume would decrease.

I'm not sure if this is the best way to implement it, but this the best way I found trying to implementing it.

### DIFF
--- a/source/funkin/backend/system/Main.hx
+++ b/source/funkin/backend/system/Main.hx
@@ -155,6 +155,17 @@ class Main extends Sprite
 		FlxG.signals.preStateSwitch.add(onStateSwitch);
 		FlxG.signals.postStateSwitch.add(onStateSwitchPost);
 
+		FlxG.signals.focusLost.add(function()
+		{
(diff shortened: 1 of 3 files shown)


NexIsDumb:
Well it would conflict with autopause or should it ignore the option

Jamextreme140:
just do a check on `Options.autopause` and it will work very well.

crowplexus:
I'm not too sure about this one, this feels like it could be made inside a mod either way…

crowplexus:
yeah nahhhh sorryyyyyy could very well work in a mod tho…

bctix:
This was made a while ago after being suggested by neo that I should make a pr after making it a soft coded addon. I totally understand if y'all don't think it's necessary to have.

crowplexus:
that's fine, i just don't think it fits the base engine (the rest of the team sort of agrees on that) though it's fine if you make an addon !

PR #367, opened 2024-08-17
Load every options file from all asset libraries

bctix:
This solves an issue I had where addons would replace the mod options.

Instead of just loading the options from the top-most library + source, it gets the options from every library.

before (only the `addon1` options are shown)
image

after (now all options from every library is shown)
image

### DIFF
--- a/source/funkin/options/OptionsMenu.hx
+++ b/source/funkin/options/OptionsMenu.hx
@@ -65,22 +65,21 @@ class OptionsMenu extends TreeMenu {
 			}
 		})]);
 
-		var xmlPath = Paths.xml("config/options");
-		for(source in [funkin.backend.assets.AssetsLibraryList.AssetSource.SOURCE, funkin.backend.assets.AssetsLibraryList.AssetSource.MODS]) {
-			if (Paths.assetsTree.existsSpecific(xmlPath, "TEXT", source)) {
+		for (i in funkin.backend.assets.ModsFolder.getLoadedMods()) {
+			var xmlPath = Paths.xml('config/options/LIB_$i');
+
+			if (Paths.assetsTree.existsSpecific(xmlPath, "TEXT")) {
 				var access:Access = null;
 				try {
-					access = new Access(Xml.parse(Paths.assetsTree.getSpecificAsset(xmlPath, "TEXT", source)));
+					access = new Access(Xml.parse(Paths.assetsTree.getSpecificAsset(xmlPath, "TEXT")));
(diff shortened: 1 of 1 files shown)


NexIsDumb:
but what if addons wanted to modify the options of that mod?

bctix:
> but what if addons wanted to modify the options of that mod?

Sorry if im not understanding, but why would an addon want to do that? 

Also, that wouldn't stop the possibility that if multiple addons want to replace a mods options, only one of those addons would show.

NexIsDumb:
> > but what if addons wanted to modify the options of that mod?
> 
> Sorry if im not understanding, but why would an addon want to do that? 
> 
> Also, that wouldn't stop the possibility that if multiple addons want to replace a mods options, only one of those addons would show.

thats what i mean, addons aren't only for overriding
that are certain addons that instead of overriding options, they **add** it

bctix:
Okay, now I know I'm misunderstanding something, leme explain why i made this pr in the first place.

Lets say i have my own mod with its own options. Then lets say i got an addon like Hud Buddies for example.

Since i added hud buddies i'd be unable to change my mods options cause HudBuddies would replace it in the options menu. I would have to remove Hud Buddies temporarily from my addons so my mods options would show. 

That's why i made this pr.

Sorry id there are any typos, im typing this out on my phone

NeeEoo:
I think like make it so if it has like the same structure? or like overrides="modname" that it would use the replaced one instead

bctix:
The last commit adds that override idea neo had. intended to be used like this.

```xml
<!DOCTYPE codename-mod-options>

<menu name="Mod Options" desc="Modify mod options here" override="modname">
	...
</menu>
```

It's intended for just mods (since mods are loaded last). I haven't tested with addons overriding addons, but I'm not sure that will happen much if at all.

NexIsDumb:
aight i think its time to consider this since i removed the examples, though wouldnt it better if the credits menu had it too for its xmls? since the behaviour is the exact same one of options’

NeeEoo:
I think another pr for the credits one would be better.

NexIsDumb:
btw, i just came back to this
i realised how dumb I was when I first saw this😭
its good, soon i'll prob make one fro the credits too

NexIsDumb:
although i think the override stuff for the xml is badly done since, what would happen if a certain xml gets read after the one that wants to override, imo its not really needed, also thanks to some stuff im doing in dev

NexIsDumb:
aight, its ready imo

PR #368, opened 2024-08-17
Added Sprite linking to FunkinSprite.hx

jasontheone111:
I coded in a way to link FunkinSprites together.
Can be used to link 2 sprite-sheets together. <- The main reason I coded this in.

:D

Usage Example:

- Create two sprites, each with different animations.
- Link sprite two to sprite one (sprite1.addLinkedSprite(sprite2, true);)
- Play anim on sprite one
- Anim does not exist in sprite one, but does exist on linked sprite (sprite two)
- Anim plays on linked sprite (Can hide parent sprite if chosen, on by default)
- Yipeee.

### DIFF
--- a/source/funkin/backend/FunkinSprite.hx
+++ b/source/funkin/backend/FunkinSprite.hx
@@ -44,6 +44,20 @@ class FunkinSprite extends FlxSkewedSprite implements IBeatReceiver implements I
 	public var animDatas:Map<String, AnimData> = [];
 
 	/**
+	 * Linked Sprite Stuff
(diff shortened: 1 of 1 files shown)


RodneyAnImaginativePerson (approved):
Very interesting. But why not instead of setting the alpha on playAnim to set it in the draw function or smth?

jasontheone111:
> Very interesting. But why not instead of setting the alpha on playAnim to set it in the draw function or smth?

Simpler

jasontheone111:
Don't pull this one, I thing I screwed something up, lemme recode it.

jasontheone111:
Made new vesrion

PR #369, opened 2024-08-17
Added Multi Sprite linking to FunkinSprite.hx (v2)

NebulaStellaNova:
I coded in a way to link FunkinSprites together.
Can be used to link 2 **or more** sprite-sheets together. <- The main reason I coded this in.
It also adds sprite tags technically, which aren't really helpful if used outside linking.

:D

Usage Example:

- Create two sprites, each with different animations. 
- Link sprite two to sprite one ( sprite1.addLinkedSprite(sprite2, true, "spriteTag"); )
- Play anim on sprite one
- Anim does not exist in sprite one, but does exist on linked sprite (sprite two) 
- Anim plays on linked sprite (Can hide parent sprite if chosen, on by default) 
- Yipeee.

### DIFF
--- a/source/funkin/backend/FunkinSprite.hx
+++ b/source/funkin/backend/FunkinSprite.hx
@@ -42,7 +42,24 @@ class FunkinSprite extends FlxSkewedSprite implements IBeatReceiver implements I
 	public var initialZoom:Float = 1;
 	public var debugMode:Bool = false;
 	public var animDatas:Map<String, AnimData> = [];
-
+	/**
+	 * Linked Sprite Stuff
+	 * defaultSelfAlpha & defaultLinkAlpha is so it can alternate between sprites if wanted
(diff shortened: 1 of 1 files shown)


RodneyAnImaginativePerson (approved):

NeeEoo:
Could you show some examples of this?

NeeEoo:
I would recommend that you check this out https://github.com/HaxeFlixel/flixel-addons/blob/dev/flixel/addons/display/FlxNestedSprite.hx, It might be better to make a FunkinNestedSprite or something

RodneyAnImaginativePerson:
> I would recommend that you check this out https://github.com/HaxeFlixel/flixel-addons/blob/dev/flixel/addons/display/FlxNestedSprite.hx, It might be better to make a FunkinNestedSprite or something

🤔 very interesting

PR #374, opened 2024-08-26
Added optional Extra Data support to ModState/ModSubState

Nikorasu-d:
## First
I dont know if this is already implemented :P, feel free to reject this if i'm wrong, feel free to comment

## What i've added?

This adds support to ModStates and ModSubstates to access entry data from a previous state

## Use Case

You will use this if you have a hub state that switchState to a handler that loads assets based on incomming data, for example:

#### WorldSelectionState.hx :
- Here you will manage the user selection and then redirect to a world
- Based on Selection you will have a switch case clausule that is going to load WorldState.hx based on user selection
- Then you will do this in any case:

```haxe 
case 0:
        FlxG.switchState(new ModState('WorldState',{
                 'worldID': 0 
        })); // This will say to state to load assets based on this.data.worldID
```

#### WorldState.hx:

- Here you will have access to the data and then you going to use it like this:

```haxe 

function create(){
	loadAssets();
}

function loadAssets(){
	trace('loading assets of World' + this.data.worldID);
}
```

## Which kind of data you can pass to this.data:

```haxe 

FlxG.switchState(new ModState('substates/tutorialSubstate',{
                'indices' : [1,2,3,4,5],
                'State' : this, //Actual State
                'function': function(){
                    trace('poto');
                },
                'String': 'HelloWorld',
                'Int': 0,
                'Float': .1,
                ... 
            }));
```

You can pass all kind of data :D, including functions, you just need to be creative

## Why you need to accept this?

Why not? jk 🅰️ XD

This allows coders to create reusable code. 
I've been working with this on my projects and it is very helpful to have fast access to data from a previous state.

### DIFF
--- a/source/funkin/backend/scripting/ModState.hx
+++ b/source/funkin/backend/scripting/ModState.hx
@@ -22,6 +22,7 @@ class ModState extends MusicBeatState {
 	public function new(_stateName:String, ?_data:Dynamic) {
 		if(_stateName != null && _stateName != lastName) {
 			lastName = _stateName;
+			data = null;
 		}
 
 		if(_data != null)
--- a/source/funkin/backend/scripting/ModState.hx
+++ b/source/funkin/backend/scripting/ModState.hx
@@ -6,16 +6,14 @@ class ModState extends MusicBeatState {
 
 	//New Constructor
 	public function new(stateName:String, ?jsonData:Dynamic = null) {
-		
 		//State Name
-		if (stateName != null){
+		if (stateName != null)
 			lastName = stateName;
-		}
-		
+
 		//Extra Data
-		if(jsonData != null){
+		if(jsonData != null)
 			data = jsonData;
-		}
+
 		super(true, lastName);
 	}
 }
--- a/source/funkin/backend/scripting/ModState.hx
+++ b/source/funkin/backend/scripting/ModState.hx
@@ -1,9 +1,24 @@
 package funkin.backend.scripting;
 
 class ModState extends MusicBeatState {
-	public static var lastName:String = null;
-	public static var data:Dynamic = null; //Use this to pass data between states (Optional)
 
+	/**
+	* Name of HScript file in assets/data/states.
+	*/
+	public static var lastName:String = null;
+	
+	/**
+	* Optional extra data.
+	*/
+	public static var data:Dynamic = null;
+
+	/**
+	* ModState Constructor.
+	* Inherits from MusicBeatState and allows the execution of an HScript from assets/data/states passed via parameters.
+	* 
+	* @param _stateName Name or path to a HScript file from assets/data/states.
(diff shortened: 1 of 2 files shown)


NexIsDumb:
oh! this looks good, but imo you should change the argument names from jsonData to data too since it can allow any kind of data (classes included for example) and also you should remove the null checks cause if the coder wants it to set it as null on pirpose and the check its still on, it can get the even previous one data (since it wont set it if the if its null because of that if condition)

NeeEoo:
Nevermind about the nullcheck

NexIsDumb:
nvm we had a better idea, thanks tho!

NeeEoo:
Could you please review the changes i pushed, and see if you are okay with them?

Nikorasu-d (commented):
Thats good, i think it is better written xd

NeeEoo:
Please use 

```
/**
 * message here
**/
public static...
```

this means it will show up in the documentation

Nikorasu-d:
> > Please use
> > ```
> > /**
> >  * message here
> > **/
> > public static...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this means it will show up in the documentation
> 
> in this way we can make the message display when hovering over the variable (in vsc for example) or in the API documentation of the engine

sure

Nikorasu-d:
> > Please use
> > ```
> > /**
> >  * message here
> > **/
> > public static...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this means it will show up in the documentation
> 
> in this way we can make the message display when hovering over the variable (in vsc for example) or in the API documentation of the engine

There it is, gonna appreciate a comment 👋

Nikorasu-d:
i've test this on a build and is a little bit buggy, so i delete the null setter, because the class already has a null for data, is not necessary

NeeEoo (changes requested):
Undo this, this has a usage to reset the stored data if the state name changes

Nikorasu-d (commented):
Think there was a problem with my specific code, gonna undo the change, sorry 🫂

NexIsDumb:
lemme test it rq and then i think we can merge :p

NexIsDumb:
lemme push some changes rq requested by Neeo

NexIsDumb:
image

yay

NexIsDumb:
and with the lastest 2 commits you can now access the vars like this in the modded state @Nikorasu-d 

image

Nikorasu-d:
> and with the lastest 2 commits you can now access the vars like this in the modded state @Nikorasu-d
> 
> image
> 
> thanks!

oh the last thing is helpful

> and with the lastest 2 commits you can now access the vars like this in the modded state @Nikorasu-d
> 
> image
> 
> thanks!
oh thanks for merging this, i'm gonna try the last thing looks good

PR #377, opened 2024-09-06
disable framerate text selection

dix-nutz:
I noticed that framerate text can be selected


NexIsDumb:
OH crap
i swear i never noticed
tysm! :D

PR #383, opened 2024-09-20
Codename Engine Mobile

mcagabe19:
NOTES:
- ~~there's some shadowed classes~~ (okay by neo)
- ~~some links needs to be changed (search MobilePorting)~~
- ~~editor warnings (search Mobile Porting Team)~~
- ~~different workflows~~
- ~~(from neo) SUtil needs to be renamed to MobileUtil~~
- ~~PsychAlphabet needs to be killed (legally can't use)~~

and idk I'll add ig and pls add `~~` if u guys figured these out


Drago-Cuven:
can't wait.

Drago-Cuven:
is there still something wrong?

NeeEoo:
we are waiting on out artist to clean up some of the assets, then also cleaning up the code a bit more, preferably we dont want buttons, but instead like real touch controls

Drago-Cuven:
> we are waiting on out artist to clean up some of the assets, then also cleaning up the code a bit more, preferably we dont want buttons, but instead like real touch controls

that makes sense, but some people probably still would want buttons,  and it'd be easier for modders.
maaaaybe between touch controls and buttons, one could be default and the other optional?

NeeEoo:
Well it would be like modes

like scroll mode which would simulate arrow presses when scrolling

gameplay mode which would be the hitboxes

custom mode which would be implemented per state

then like button mode

Drago-Cuven:
ah
ok

NexIsDumb:
okie im just making this comment to remind to myself or anyone else that wants to help for the pr

this are all the missing things or bugs i found:

- every shader crashes:




- Need to adjust secondary threads stuff, maybe a queue? cuz github icons arent downloading all (or it might be cuz rn i have a slow connection since im in class):


- for some reason the test song doesnt load?? just the dialogue:

NexIsDumb:
oh! okay!
just found out that test didnt load cuz i was trying to play it on hard while it only has normal, altho its odd? the engine should recognise if the song doesnt have certain difficulties and not allowing you to select them; i guess thats bugged on mobile; i'll check

NexIsDumb:
oh nvfm test crashes, a tween crash it seems

mcagabe19:
shader is indeed broken, idk about test and github thing due slow internet

PR #388, opened 2024-09-29
fixes to character offsets/indices (Not meant to reflect base game)

NyxieFemboy:
These changes are purely to make the pose offsets and animations feel more fluid and consistent, these will very much not fit what they look like in base game and are purely based on what I feel looks best in-game


NexIsDumb:
these all look good, although, can you recheck parents christmas? the throne slides lmao

NexIsDumb:
oh and could u add -loop on the idle of xmas parents too?

PR #390, opened 2024-10-01
fixed Beats Per Measure not being updated based on sectionBeats

NyxieFemboy:
i noticed that sectionBeats was just completely unused which meant that it was never updated beatsPerMeasure and in turn, all measures were treated like they had 4 beats regardless of whether or not that was true.


NexIsDumb:
Contained in #397

PR #414, opened 2024-10-23
Update TitleState.hx

NebulaStellaNova:
Title Screen More Modifiable Via Script


NexIsDumb:
makes no sense, you can still do it through script even if they arent public

PR #415, opened 2024-10-24
Added Timer functionality to PlayState.hx

THEkatinamicrowave:
Images are made but still need to be added


KadePleaseHelpMe:
how do I add the images and stuff

KadePleaseHelpMe:
Base:
base
Actual coded thing:
actualcodedthing
Decor-Clock:
decor_clock
Decor-Clock Hand
decor_hand

NexIsDumb:
even if this looks very sick, i dont think we can add it on the base engine like this unfortunately: it'd break past mods, although you can make a script about this and publish in the discord server cuz it really looks sick!

PR #418, opened 2024-10-29
Fix dialogueLine next sound getting from assets/music instead of assets/sounds

joaquinLOL123:
Just a fix for an error I noticed while working on a mod where if the dialogue next sound is set by a line in dialogue.xml, it would try to grab from the music folder instead of sounds.


NexIsDumb:
oh hell?? I never noticed
I'll check better when i return home

NexIsDumb:
oh crap youre right

PR #425, opened 2024-11-03
remove `sudo` requirement when launching on the CLI under Linux systems

crowplexus:
release my family please.


NexIsDumb:
> release my family please.

duolingo v2???

PR #432, opened 2024-11-05
Live charting in Charter

THEkatinamicrowave:
I already made en enhancement issue about this, so I'll copy and paste it here:

Ok so hear me out.
You're trying to make a chart for a song you made, except you decided to make a one-shot mod so your song is like 8 minutes long. Charting usually takes ``time of song`` * 185323, so this idea should make charting quicker.

Using a keybind like Ctrl + R, the charter will prompt you to click a strumline to select it for charting. Once selected, you can press space to begin a countdown, starting at the beginning of your chart (Ctrl + Shift + R to start at current position). During this recording, any keybinds that you press will be converted to notes (don't worry, they'll be snapped to whatever grid you set (the one that's 16 by default). Once you press Ctrl + R again, the recording is stopped and you can go in to manually edit whatever charting you didn't like.

Thought it was a cool idea and it would WAY speed up charting. Only problem I can see with this is having input delay, but you can probably easily offset the chart based on Options.songOffset.


KadePleaseHelpMe:
Why aren't the notes appearing onscreen
Well I mean they are but like if I just ``add()`` instead of ``notesGroup.add()`` it doesn't work like a charternote (I've tried so hard to read the api)

KadePleaseHelpMe:
@NeeEoo it's not letting me auto-update the stuff how do I make my branch utd

THEkatinamicrowave:
YOOOO IT LIKE SOMEWHAT WORKS NOW

THEkatinamicrowave:
pointless; also literally has branch conflicts

PR #436, opened 2024-11-08
Chromatic Aberration shader 

THEkatinamicrowave:
I unbroke the chromatic aberration shader


NeeEoo:
This is intentional you have to do

redOff = [value, 0];

KadePleaseHelpMe:
oh ok thank you

PR #438, opened 2024-11-10
add a variable on PlayState to hide the "COMBO" Sprite

crowplexus:
also added another to control when the combo numbers should show up based on your combo counter

also makes it more in-line with new base game


crowplexus:
brb lol!

PR #463, opened 2024-11-29
GitHub API now uses per_page argument to show more of requested data

FuroYT:


NexIsDumb:
Imo if the max is 100, isnt it better to actually place 100 as default?

NeeEoo:
nah i would use 30 and use the ?page=NUM when the user scrolls down enough

NexIsDumb:
fair but the api is rate limited

NeeEoo:
right- i forgot that

NexIsDumb:
> We could have a button that says
> 
> Load More (120s...)

can you even decide the start point? otherwhise youd download just the same list

NeeEoo:
Yes you can, check the api

NexIsDumb:
sweet

PR #501, opened 2024-12-16
Auto `FlxG.save.bind` when switching mods (so no mods override other mod saves and removes bloat)

ItsLJcool:
There probably is a better way to do it than just `FlxG.save.bind("save-default", 'CodenameEngine/${ModsFolder.currentModFolder.trim()}');` in `MainState`, though when people download the newest action build they will lose all their saves for mods (example: YTP Invastion) because they relied on `FlxG.save` from CodenameEngine saves.

This won't override the `Options.__save` because they differ from base Flixel's FlxSave, and now I can actually do something niche with this feature that im going to be adding :eyes:

### DIFF
--- a/source/funkin/backend/system/MainState.hx
+++ b/source/funkin/backend/system/MainState.hx
@@ -68,5 +68,10 @@ class MainState extends FlxState {
 		}
 
 		CoolUtil.safeAddAttributes('./.temp/', NativeAPI.FileAttribute.HIDDEN);
+
+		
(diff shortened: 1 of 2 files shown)


NexIsDumb:
okay honestly this is one of the dev stuff we were thinking of, I might merge this very soon

NexIsDumb:
nvfm, lets burn this with forks!!!! (we talked in private)

PR #507, opened 2024-12-17
Add Customizable Debug Key

THEkatinamicrowave:
using the unused (tmk) "cheat" key


KadePleaseHelpMe:
crazy hamburger

KadePleaseHelpMe:
also I playtested it this works

ItsLJcool:
I don't think they will accept this pr since you can just do what Zoro did in his mod lol
image
image

but yeah there should be a way to allow custom keybinds, you should do that instead of repurpousing the Cheat key (also its not entirely unused, the switch can hit that key)

PR #512, opened 2024-12-29
inequality sign in FlxAnimate

DixNutzzz:
Is it just me, or is it the wrong sign?


Raltyro:
Yeah
though in that case just remove these 2 codes, since openfl default blendmode are already normal

NeeEoo:
Oh my god im stupid

PR #513, opened 2024-12-30
[ENHANCEMENT] Prevent accidental selecting when tabbing back in

TechnikTil:
This PR prevents accidental selection when tabbing back in. It does this by adding a timeout for a quarter of a second.


NexIsDumb:
now its accessible to eevryone as @NeeEoo requested

PR #527, opened 2025-01-12
Adding erect stages

KadePleaseHelpMe:
Pull made to add the erect stages to CNE. Sort of just a personal test of my ability to port V-Slice stuff to CNE.


NeeEoo:
This should be in https://github.com/CodenameCrew/V-Slice-Modpack

PR #542, opened 2025-01-31
Change "Press ENTER" to "Press any key"

SpareEnderboy:
og description: no but a tin can
og title: can a match box
reference: q2 of the impossible quiz and corresponding answer


SpareEnderboy:
image

PR #546, opened 2025-02-03
Remove chart editor acess restriction

SorbetLover:


ItsLJcool:
Its there for a reason, if i where the devs i prob wouldn't pull the pr

SrtHero278:
Yeah......
The charter _**is**_ a debug thing, so it makes sense as to why it's in the `canAccessDebugMenus` conditional.

NeeEoo:
Tho i once tried enabling the vars using code
but didnt seem to work

SorbetLover:
at least add an option to take the restriction off, its annoying finding a bug on the song and having to exit, go to chart selection and then the editor

PR #547, opened 2025-02-03
Noteskin fixes :3

GlaveJayce:
So yeah i fixed the stuff, nothing truely gamebreaking but mods that have accounted for this error may now have the left strum as the hold end

but this now makes All noteskins that arent defined within its own code (such as notes for mines etc) fixed


usb-port-2:
Make it detect both the fixed and original version otherwise this will break ALL previous mods

GlaveJayce:
Ill try!

GlaveJayce:
image
Well, to be quite honest, I have not been able to find a solution, but its not a true gamebreaker

usb-port-2:
Psych managed to be compatible with both

PR #548, opened 2025-02-08
Multikey support

THEkatinamicrowave:
easiest pr ever
also untested so don't blame me if it's broken


NeeEoo:
Theres already multikey support

PR #549, opened 2025-02-09
Make the level info text in `PauseSubState` accessible to scripts

charlesisfeline:
just what the title says. i want to hide the difficulty text in the pause menu, pls just let me...


NexIsDumb:
looks fine, lemme actually edit smth little i was planninng

NexIsDumb:
there we go

PR #552, opened 2025-02-15
openalsoft audio fixes

cyn0x8:
just a mirror pr of https://github.com/FunkinCrew/Funkin/pull/3318 and https://github.com/ShadowMario/FNF-PsychEngine/pull/15536

changes some openalsoft config settings via a config file to prevent it from doing virtualization when it detects your device as "headphones", which ruins stereo fullness, especially in lower frequencies

also as a bonus this config disables the limiter + slightly improves audio resampling

sooo with this change the audio should sound just like as if you opened it with a media player app (original quality!!!)

---

<details><summary>example vids (left side is the new audio)</summary>

<br>

https://github.com/user-attachments/assets/171259a1-928d-4389-9a62-afc2c3376665

https://github.com/user-attachments/assets/0cf5511d-8f11-4b5c-bc55-2b0aabd345ab

</details>

these arent the best recordings since rn im using a realtek driver instead of my dac so the eq is kinda fucked up in obs... but the config does work!! you can hear the stereo is more full

### DIFF
--- a/alsoft.txt
+++ b/alsoft.txt
@@ -1,0 +1,14 @@
+[general]
+sample-type=float32
+stereo-mode=speakers
+stereo-encoding=panpot
+hrtf=false
+cf_level=0
+resampler=fast_bsinc24
+front-stablizer=false
+output-limiter=false
+volume-adjust=0
+[decoder]
+hq-mode=false
+distance-comp=false
+nfc=false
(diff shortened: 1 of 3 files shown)


NexIsDumb:
holy shit this might be peak
gotta wait for @NeeEoo to become available again though to see if hes fine with this aswell
hes gonna be unavailable for like 1 week or smth, have patience 
but rest assure that we're gonna most prob merge this

NexIsDumb:
okay nvm this actually works well, i dont think i need neeo's confirmation, thanks!

PR #572, opened 2025-03-07
add indicator on how to open debug tools

al-ex427:
If you can open debug tools, make it visible on how to open


r6915ee:
Considering the fact that it's already a standard to use `7` as the key for opening debug menus in FNF engines, I'm not sure how useful this would be personally. Most people using this engine come from, for example, Psych.

NeeEoo:
While i kinda agree it feels useful
You can only know about it if you know it, like no one would guess it.
And we only have it mentioned on the website
And not everyone who uses codename engine used psych

r6915ee:
I know, I'm just saying that it's not really expected to have it directly outlined within the engine itself considering how the key has been handled.

dotaxel:
verison 😭🙏

moxie-coder:
> verison 😭🙏

IMG_3424

PR #599, opened 2025-04-05
i can code too!

SpareEnderboy:
press any key is a reality bitches


NeeEoo:
Your new line handling on your system causes commits which are very annoying to differentiate.
If you use github desktop it should handle this automatically

SpareEnderboy:
> Your new line handling on your system causes commits which are very annoying to differentiate. If you use github desktop it should handle this automatically

i use vs code.
i do not want to configure git.

SpareEnderboy:
trying to commit PROPERLY

r6915ee:
> i use vs code.

Visual Studio Code allows you to change the newline handling through this button at the bottom toolbar on a default layout:

image

PR #608, opened 2025-04-08
Winter Horrorland Linux Fix (Attempt 2)

EfanzeDev:
* Fixed Winter Horrorland searching for char Monster-christmas (which technically doesn't exist on platforms other than windows due to capitalisation fuckery), this time using the original chart format.


NexIsDumb:
ah fair enough ty!

PR #615, opened 2025-04-16
Script call when setting Downscroll.

rodney528:
Let's say your HUD needed some changes when downscroll was enabled. This script call would help you be able to still switch downscroll on the fly without fucking up your HUD in the process!

Ik the way I explained it is a bit dumb, but this would be pretty cool.

### DIFF
--- a/source/funkin/game/PlayState.hx
+++ b/source/funkin/game/PlayState.hx
@@ -138,7 +138,8 @@ class PlayState extends MusicBeatState
 	public var downscroll(get, set):Bool;
 
 	@:dox(hide) private function set_downscroll(v:Bool) {
-		scripts.call('onDownscrollSet', [v]);
+		if (camHUD.downscroll != v)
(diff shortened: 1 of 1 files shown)


NeeEoo:
Id say make it only run if(v != camHUD.downscroll)

FuroYT:
I don’t believe that is necessary since downscroll can only be changed by normal means in the options menu which unloads PlayState

dotaxel:
this is totally unnecessary
u can do the same with some lines of code, like:
```haxe

var lastDws = downscroll;

function update()
{
    if (lastDws != downscroll)
    {
      // code when downscroll state changes
    }
    lastDws = downscroll;
}
```
:]

NexIsDumb:
> this is totally unnecessary
> u can do the same with some lines of code

i honestly feel the same cuz script calls may be heavy sometimes based on scripts, so i rather reduce the use of them for unnecessary stuff

PR #617, opened 2025-04-21
Add a new function for CoolUtil

VapireMox:
abab
...
Content: 
Safely retrieve all files in the specified directory and subdirectories

### DIFF
--- a/source/funkin/backend/utils/CoolUtil.hx
+++ b/source/funkin/backend/utils/CoolUtil.hx
@@ -80,6 +80,46 @@ class CoolUtil
 		#end
 	}
 
+	@:noUsing public inline static function safeGetAllFiles(originPath:String, reservePath:Bool = false, stayEmptyDirectory = true, showErrorBox:Bool = false):Array<String> {
+		#if sys
+		var files:Array<String> = [];
+		originPath = Path.addTrailingSlash(originPath);
+
+		var errorMessages:Array<String> = [];
+		function traverse(currentPath:String) {
+			try {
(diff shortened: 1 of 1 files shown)


r6915ee:
Shouldn't you still be able to amend commits in the event that it's related to the prior one and doesn't contain many additions?

VapireMox:
> Shouldn't you still be able to amend commits in the event that it's related to the prior one and doesn't contain many additions?

What?

r6915ee:
> What?

The second commit removes the second version of the function. This can be easily done with one commit instead by amending the changes to the latest one.

VapireMox:
> > What?
> 
> The second commit removes the second version of the function. This can be easily done with one commit instead by amending the changes to the latest one.

I know, but it's unavoidable. I forgot to delete the additional function

r6915ee:
> I know, but it's unavoidable. I forgot to delete the additional function

Amending usually appends the changes directly. For example, if line A and B were to be added and an amend is made to remove line B, only line A should remain in the commit AFTER it has been committed.

VapireMox:
> > I know, but it's unavoidable. I forgot to delete the additional function
> 
> Amending usually appends the changes directly. For example, if line A and B were to be added and an amend is made to remove line B, only line A should remain in the commit AFTER it has been committed.

So this pull request has been invalidated?

VapireMox:
If that's the case, it's good because I found some other errors inside

r6915ee:
> So this pull request has been invalidated?

Not really, it's just good practice to amend in such a case.

PR #619, opened 2025-04-24
More event params !!!!!!!!!!!!! yipeee!!!!!!!!!!1

THEkatinamicrowave:
- Adds and angle parameter to Add Camera Zoom event
- You can have negative camera zoom (this should have always existed)
- Quarter beats are supported in Camera Modulo Change

I assume this would fuck up old mods so this is still WIP
also can someone please tip me off as to why this is commented out by default
image

### DIFF
--- a/source/funkin/backend/chart/EventsData.hx
+++ b/source/funkin/backend/chart/EventsData.hx
@@ -19,11 +19,12 @@ class EventsData {
 		"Camera Movement" => [{name: "Camera Target", type: TStrumLine, defValue: 0}],
 		"Add Camera Zoom" => [
 			{name: "Amount", type: TFloat(-10, 10, 0.01, 2), defValue: 0.05},
+			{name: "Angle", type: TFloat(-180, 180, 0.01, 2), defValue: 0},
 			{name: "Camera", type: TDropDown(['camGame', 'camHUD']), defValue: "camGame"}
 		],
 		"Camera Modulo Change" => [
-			{name: "Modulo Interval (Beats)", type: TInt(1, 9999999, 1), defValue: 4},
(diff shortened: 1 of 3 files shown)


FuroYT:
why is there angle stuff in a zoom event??

THEkatinamicrowave:
> why is there angle stuff in a zoom event??

it would be so awesome
it would be so cool
just think about it
you have a cool punch-in effect with a lil dutch angle for extra effect

FuroYT:
no it won’t??
plus you can easily script it yourself so there isn’t a use for that

chronicsilly:
why negative camera zooms...

THEkatinamicrowave:
> no it won’t?? plus you can easily script it yourself so there isn’t a use for that

the coolness is fully subjective on your part

THEkatinamicrowave:
> why negative camera zooms...

why positive camera zooms
hell camera zooms aren't even used in any of the vanilla charts so why make it an event at all

r6915ee:
> also I'm porting a mod from Psych to CNE which has negative zooms and it's kind of a pain in the ass

Why would any mod, regardless of engine, require negative camera zooms in the first place? That seems somewhat unnecessary to add functionality for that, considering the fact that it's not something you would use often; heck, I don't even *know* of a single mod that does this.

THEkatinamicrowave:
> why do we need negative camera zoom

punch out zooms.

> I don't know a single mod that does this

Funkin' Family Rush 3.0 - Sticks

NexIsDumb:
sorry but as the others say this looks a bit of a useless bloat, besides many code lines breaks a tons of mods, starting with editing the parameters of the cam zoom event for example

PR #642, opened 2025-05-30
Fix wmic

FuroYT:
Fixes the engine getting your cpu name and ram type for the F3 menu and also adds a new class to look into the windows registry (returns `null` and `false` on non windows platforms)
image
(_i dont really care that im leaking my specs_)

Before any regards the `RegistryUtil` class only has access to what the reg command has on windows so it cant access admin based keys if you don't run the program as admin

**Edit**
Now actually uses the untranslated string of the windows version
image

### DIFF
--- a/source/funkin/backend/utils/MemoryUtil.hx
+++ b/source/funkin/backend/utils/MemoryUtil.hx
@@ -121,7 +121,7 @@ class MemoryUtil {
 		];
 		var memoryOutput:Int = -1;
 
-		var process = new HiddenProcess("wmic", ["memorychip", "get", "SMBIOSMemoryType"]);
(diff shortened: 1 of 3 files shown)


NexIsDumb:
oh wait peak

source/funkin/backend/utils/MemoryUtil.hx
```diff
@@ -121,7 +121,7 @@ class MemoryUtil {
 		];
 		var memoryOutput:Int = -1;
 
-		var process = new HiddenProcess("wmic", ["memorychip", "get", "SMBIOSMemoryType"]);
+		var process = new HiddenProcess("powershell", ["-Command", "Get-CimInstance Win32_PhysicalMemory | Select-Object -ExpandProperty SMBIOSMemoryType" ]);
```
NexIsDumb, line 124: are we sure this wont be blocked by antiviruses like the one for setting hidden files long ago?
cuz i mean youre starting a powershell hidden process

PR #654, opened 2025-06-17
Add noExtension flag to getFolderContent for cleaner Paths usage

HeroEyad:
I think this is annoying every time I have to do `Path.withoutExtension()` manually just so I can use `Paths.image()` when pushing folder contents to an array.

This PR adds a new `noExtension:Bool = false` argument to `Paths.getFolderContent()` to simplify that process. Internally, it uses `Path.withoutExtension()` to return filenames without their extensions when the flag is set.

**Why this helps:**

* Makes folder-to-path workflows easier, especially with `Paths.image()`, `Paths.font()`, etc. (this would work with easter eggs and such)
* Cuts down repetitive code (`Path.withoutExtension(...)`).
* Keeps default behavior intact (`noExtension = false`).

**Example usage:**

```haxe
var iconFiles:Array<String> = [];
var icons = Paths.getFolderContent("images/icons", false, false, true); // noExtension = true

for (icon in icons) {
    iconFiles.push(icon);
}

var spr = new FunkinSprite(0,0,Paths.image(FlxG.random.getObject(iconFiles)))
add(spr);
// or make a better use to it idk.
```

Before: you'd have to strip extensions manually.

Now: just set `noExtension` to true and you're done.


HeroEyad:
thank you furo pls dont kill me

PR #655, opened 2025-06-18
Credits Blacklisting (User IDS)

Hackx2:
Pretty self-explanatory... 

How to get a user's ID:
https://api.github.com/users/USERNAME_GOES_HERE_DUH

:D

### DIFF
--- a/source/funkin/menus/credits/CreditsCodename.hx
+++ b/source/funkin/menus/credits/CreditsCodename.hx
@@ -17,6 +17,8 @@ class CreditsCodename extends funkin.options.OptionsScreen {
 	public var minContrCol:FlxColor = 0xFFB4A7DA;
 	public var contribFormats:Array<FlxTextFormatMarkerPair> = [];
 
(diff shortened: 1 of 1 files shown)


r6915ee:
What would this be needed for? Besides, there's already the commit history that complicates these kinds of things; that is its own can of worms that can possibly impact this, especially with things like integrity.

chronicsilly:
"I'll add a... credits blacklist"
"How original"
"And add quirky :3 comments in the code"
"Daring today, aren't we?"

(BTW this isn't going to be a thing because Fushi thinks deleting history is history, like we didn't erase the Nazis from the world's history we learnt from it and study it in schools :P)

hackx2:
True! 

I never really thought about that way. 
Sorry..

(I take pride in my comments 😭)

chronicsilly:
> True!
> 
> I never really thought about that way. Sorry..
> 
> (I take pride in my comments 😭)

I thought the comments were nice its just like 3 other people have made this exact pull request in the past </3

PR #662, opened 2025-06-27
Week6 pixel fix

ItsLJcool:
Human... I remember you
<r>fixing those Week 6 antialiasing issues<r>

me when <g>LJ<g> scares the people using his <y>Auto Action Build Updater<y>


NexIsDumb:
lemme just check how it works in latest

PR #663, opened 2025-06-28
Fix Inconsistency in Sustain Input

Raltyro:
Makes it so Player Input on Sustain as the same as how its supposed to be or how the cpu does it

Before: 
Player hits the sustains and sustain animation ending too early

https://github.com/user-attachments/assets/4763fb10-51e7-4435-b819-faf0a52c230e

After: 
Player hits the sustain and sustain animation in time

https://github.com/user-attachments/assets/4b997151-f170-420b-bc60-faf24526a0e1

### DIFF
--- a/source/funkin/game/Note.hx
+++ b/source/funkin/game/Note.hx
@@ -276,30 +276,22 @@ class Note extends FlxSprite
 	public function updateSustain(strum:Strum) {
 		var scrollSpeed = strum.getScrollSpeed(this);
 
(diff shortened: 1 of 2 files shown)


crowplexus:
technically solved internally because inputs are meant to be rewritten but *until* I can compile the game to test, I'll just give this tags

Raltyro:
Then, can we have this temporarily? since it's solved internally because the inputs are rewritten in dev branches?

PR #665, opened 2025-06-30
Fix scaling bug on FlxView3D on High DPI displays

kitzsh:
Fixes a bug in FlxView3D where the underlying View3D is larger than the BitmapData it renders to when "allow-high-dpi" is enabled, only showing the top left-hand corner of the view.

### DIFF
--- a/source/flx3d/FlxView3D.hx
+++ b/source/flx3d/FlxView3D.hx
@@ -5,6 +5,7 @@ import away3d.containers.View3D;
 import away3d.library.assets.IAsset;
 import flixel.FlxG;
 import openfl.display.BitmapData;
+import openfl.system.Capabilities;
 #end
 import flixel.FlxSprite;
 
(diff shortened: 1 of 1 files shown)


kitzsh:
Needs more testing actually

kitzsh:
Tested it with a few different resolutions, should be all clear

crowplexus:
OH SO THAT'S WHY????

kitzsh:
yup

TechnikTil (approved):
ohhhh
so thats why

NexIsDumb:
was this actually tested enough?

Raltyro:
There has to be another issue relating to dpi, because texts just gets messed up if the current game resolution isnt in default

MRK2009:
It got conflicted.

kitzsh:
> It got conflicted.

Doesn't really matter, this PR is no longer needed

MRK2009:
Oh okay nvm then

Issue #14, opened 2022-12-14
Compiling help: VLC/MP4Handler Errors


te-agma-at:
[weed]: <> (FILL THIS ISSUE THING OUT AS MUCH AS POSSIBLE)
[weed]: <> (OR ELSE YOUR ISSUE WILL BE LESS LIKELY TO BE SOLVED!)
[weed]: <> (DO NOT POST ABOUT ISSUES FROM OTHER FNF MOD ENGINES! I CANNOT AND PROBABLY WON'T SOLVE THOSE!)
[weed]: <> (GO TO THEIR RESPECTIVE GITHUB ISSUES AND REPORT THEM THERE LOL!)

#### Please check for duplicates or similar compiler issues by filtering for 'compiler help'

[weed]: <> (Put an X in the [ ] thingies to fill out checkbox!)
[weed]: <> (something like [x] pretty much, don't screw up or you will look stupid)

- [X ] Windows
- [ ] Mac
- [ ] Linux
- [ ] HTML5

## Please describe your issue. Provide extensive detail and images if possible.

Error message output while compiling:
source/funkin/cutscenes/MP4Handler.hx:4: characters 8-28 : Type not found : vlc.bitmap.VlcBitmap
source/funkin/cutscenes/MP4Handler.hx:17: characters 20-29 : Type not found : VlcBitmap
source/funkin/game/PlayState.hx:66: lines 66-1452 : ... Defined in this class
source/funkin/cutscenes/VideoCutscene.hx:31: characters 17-33 : funkin.cutscenes.MP4Handler does not have a constructor
source/funkin/cutscenes/VideoCutscene.hx:33: characters 15-26 : funkin.cutscenes.MP4Handler has no field canvasWidth
source/funkin/cutscenes/VideoCutscene.hx:34: characters 15-27 : funkin.cutscenes.MP4Handler has no field canvasHeight
source/funkin/cutscenes/VideoCutscene.hx:41: characters 15-22 : funkin.cutscenes.MP4Handler has no field playMP4

VlcBitmap type might need to be added or MP4Handler fixed.

te-agma-at:
Marked as closed. Solution has been found:

- Go to Command Prompt
- Install hxCodec by running `haxelib install hxCodec 2.5.1`

PR #666, opened 2025-06-30
FINALLY fixed the issue with GlobalState blowing up when pressing F5.

ItsLJcool:
This happened due to the GlobalScript in `assets/data/global.hx` containing the code to reset the state being non functional, I wont explain in depth but basically spamming F5 caused it to blow up GlobalScripts.

I removed ALL instances of the Source code using `FlxG.keys.justPressed.F5` to a new Control Key `DEBUG_RELOAD`
So now you can set a custom keybind to reload states in the Options.

To reload GlobalScripts use `SHIFT` + `DEBUG_RELOAD`.

CURRENTLY THERE IS AN ISSUE, BUT IT'S [FIXED IN THIS PR](https://github.com/CodenameCrew/cne-flixel/pull/14)
once that is merged, this shouldn't cause any issues.

The issue?
It will reload twice, due to an oversight with updating the input AFTER `preUpdate`... but elapsed is updated before it :sob:

anyways <r>RED JUMPSCARE!!<r>


crowplexus:
I need to test this further but I'll let the rest of the team know

ItsLJcool:
oh also btw if you guys could get F5 to work for resetting substates I'd appreciate it, but I couldn't figure out how to reset the current open substate so rip

NexIsDumb:
I mean, i think at this point we can remove the global script from base assets
i'll test this and merge it in the future if nobody does it before, tyy

NexIsDumb:
also like @SenDoesStuff was suggesting, this pr makes me more into the idea of having a developer mode in the options, for like accessing editors, some keybinds and more

ItsLJcool:
No your right, but I just kept the file in so people can notice the file structure, and if we need to add anything there lol

> I mean, i think at this point we can remove the global script from base assets i'll test this and merge it in the future if nobody does it before, tyy

NexIsDumb:
> No your right, but I just kept the file in so people can notice the file structure, and if we need to add anything there lol

nah imo just let em use the website for that

ItsLJcool:
I was planning on making a pr on adding a Toolbox with the debug keys packaged in it, but I'm not sure if I really want to do that, but its been in my mind.

> also like @SenDoesStuff was suggesting, this pr makes me more into the idea of having a developer mode in the options, for like accessing editors, some keybinds and more

NexIsDumb:
> I was planning on making a pr on adding a Toolbox with the debug keys packaged in it, but I'm not sure if I really want to do that, but its been in my mind.
> 
> > also like @SenDoesStuff was suggesting, this pr makes me more into the idea of having a developer mode in the options, for like accessing editors, some keybinds and more

i'll see it myself since i'll do it in dev, dw

ItsLJcool:
[#14](https://github.com/CodenameCrew/cne-flixel/pull/14) cne flixel branch merged, testing new commit

ItsLJcool:
ok pr should be good to go.

Currently Reloading SubStates won't work because of overriding new functions, so it would be a weird fix to make it work, and that fix will break or be annoying to maintain, so until we can find a better one, reloading States should just be fine for now.

NexIsDumb:
lemme check

ItsLJcool:
ooh wait @NexIsDumb should we quickly add a keybind for console debug?

ItsLJcool:
or whole other pr?

NexIsDumb:
> ooh wait @NexIsDumb should we quickly add a keybind for console debug?

nah i'll do smth like in dev as I've already planned

PR #669, opened 2025-06-30
Update FreeplayState.hx

NebulaStellaNova:


crowplexus:
why though?

NexIsDumb:
there was a reason why the parsed field is there, and reason is to avoid running that on update everytime, why are you changing that?

TechnikTil:
If `parsedColor` and `FlxColor.fromString(color)` are the same thing, just `parsedColor` is parsed beforehand, why is this needed exactly?

ItsLJcool:
uh I believe parsedColor was used because no one thought about just parsing it with `FlxColor` and it was ease of use. tbh idek why it existed lmao

in this case, if we merge this pr we would need to replace `parsedColor`, and its only purpouse was to save the result into the instance, It was also noted to not even export to the json but to just be a data holder
image

NexIsDumb:
wait lemme check smth

NexIsDumb:
nnope, its better not to merge this since everytime the meta loads parsedcolor gets used automatic
its just a way to reduce calculus and optimise
immagine

...buuuuutt, you gave me an idea, its really annoying indeed have two values, so why not making color already an FlxColor?
IT COULD break few scripts but imo its really worth it

NexIsDumb:
i'll work on it rn

NexIsDumb:
it _should_ be good

PR #674, opened 2025-07-02
[BUGFIX] use `haxelib run lime` for compiling in commandline

TechnikTil:
## What does this PR do?
Before, `commandline` would work by using whichever `lime` was used in `haxelib run lime setup`. This though, could cause some compatibility issues in the long run.

This PR makes it so `commandline` will now force people to use the `cne setup` (aka `hmm`) version of lime for commands, so these issues will never occur.

you might've been using a different version of lime for compiling
you just didn't know it yet!


NexIsDumb:
oh wait this is good

PR #687, opened 2025-07-15
Why the fuck

chronicsilly:
this is really unnecessary and confuses people and not everyone has a numpad so like

who did this and why

image


chronicsilly:
not me spending 20 minutes checking if my keyboard keys worked properly before i went to settings and saw someone had set volume to be controlled by the non-existant numpad

PR #703, opened 2025-08-05
[LANG] Add Portuguese

Bitto1070:
lots of alphabet characters missing and the stage editor changes the name of the window to "F" somehow but
markiplier

### DIFF
--- a/assets/data/alphabet/bold.xml
+++ b/assets/data/alphabet/bold.xml
@@ -97,7 +97,7 @@
 	</composite>
 	<composite char="à">
 		<component anim="character-a0" x="0" y="0"/>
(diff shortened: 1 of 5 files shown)


Verwex:
I would need someone that knows portuguese to verify this but so far it looks good. I like how some words that can't be translated as easily are put in quotation marks. That's nice. Though I have a couple of things to say:

- A couple control options go off the path, but then again english has that same issue
- If "Ghost Tapping" is in quotations, then why isn't Downscroll
- Streaming Music and Streaming Vocals need to be shortened somehow
- A bunch of stage editor strings are not translated, like {stageCreationScreen.stage-name} or {stageInfoScreen.stageName}

So far though, it's looking good in my books 👍

PR #710, opened 2025-08-07
added needsVoices to the metadata screen

KolzeYT:
needs translations to italian and spanish
(i really need this for the vslice pack)


Raltyro:
Was thinking about just deprecating needsVoices and adds a voicesPath something, so eh

PR #712, opened 2025-08-07
wouldn't return null, would just error

rodney528:


FuroYT:
Can you explain your changes so I know what you did there cus that looks interesting

rodney528:
> Can you explain your changes so I know what you did there cus that looks interesting 

I'm guessing the strumLines are killed and not destroyed because when testing something, I went from 4 to 2 strumLines but Charter.strumLines.length would still return 4. So I tested doing forEachAlive and it became accurate to the actual amount of strumLines present.

Raltyro:
Technically fixed by updating the length of the strumLines group (recent commit)

PR #718, opened 2025-08-08
Options XML crash fix

rodney528:


source/funkin/options/OptionsMenu.hx
```diff
@@ -151,6 +151,11 @@ class OptionsMenu extends TreeMenu {
 		var options:Array<FlxSprite> = [];
 
 		for(node in xml.elements) {
+			switch(node.name) { // literally just so you don't need "name" because there is no need for this to require it 😭  - RodneyFurball
+				case 'separator':
+					options.push(new Separator(Std.parseInt(node.has.height ? node.att.height : 67)));
```
r6915ee, line 156: 67 is already an integer. As such, Haxe refuses to compile this, because the `x` parameter in `Std.parseInt` can [**only be a string**](https://api.haxe.org/v/4.2.5/Std.html#parseInt).

r6915ee (changes requested):
The code cannot compile, due to **mismatched types** on line 156.

source/funkin/options/OptionsMenu.hx
```diff
@@ -153,7 +153,7 @@ class OptionsMenu extends TreeMenu {
 		for(node in xml.elements) {
 			switch(node.name) { // literally just so you don't need "name" because there is no need for this to require it 😭  - RodneyFurball
 				case 'separator':
-					options.push(new Separator(Std.parseInt(node.has.height ? node.att.height : 67)));
+					options.push(new Separator(Std.parseFloat(node.has.height ? node.att.height : 67)));
```
r6915ee, line 156: The same issue from before remains in that mismatched types are passed. What is perhaps the main issue is the 67.

What `Std.parseInt` and `Std.parseFloat` do is that they parse from a string, and *only a string*, and then return the data they find from the string and return that data as their respective return type. The problem here is that an integer is a possible parameter being passed to this function. **Because both of these functions can only take in a string as the parameter and a possible value that can be passed to this function is an integer, this will cause the compiler to quit.**

My suggestion would be, if we want to make things simple, to instead make the 67 a string. So, a revised version of this line that not only is simple but still avoids the compiler error can be:

```haxe
options.push(new Separator(Std.parseFloat(node.has.height ? node.att.height : "67")));
```

r6915ee (changes requested):

r6915ee (approved):
I'm able to compile now; it seems to be much more stable, at least from what I can get this pull request is trying to do.

rodney528:
> I'm able to compile now; it seems to be much more stable, at least from what I can get this pull request is trying to do.

Yeah, it fixes "forId" needing to exist even though the argument is set to be optional. Similarly "height" had a default value for the argument so I added a check for that too. As well as made it so "Separator" didn't need a name since there's no point in it having one.

r6915ee:
> Yeah, it fixes "forId" needing to exist even though the argument is set to be optional. Similarly "height" had a default value for the argument so I added a check for that too. As well as made it so "Separator" didn't need a name since there's no point in it having one.

Knowing how weird Haxe is with errors and the fact that it can be very unfriendly with them, that's unfortunately expected with the previous setup, so I'm at least glad this pull request is up.

Raltyro:
wait
what did i do wrong why is it not merged yet

Raltyro:
are you kidding me squash and merge and squashing merging in fork is different thing

PR #722, opened 2025-08-09
Disable filefilter in UIFileExplorer on macOS

lasystuff:
its broken on mac


lasystuff:
I'm really dumbass it wasn't

PR #739, opened 2025-08-13
cam movement keybind quanted & added a local keybind

KolzeYT:
just that


FuroYT:
me thinks you should add an argument to choose whenever or not you snap the event or not (ie. pressing SHIFT disables the snap)

NexIsDumb:
i couldnt test it myself but im pretty sure its fine

PR #742, opened 2025-08-14
onPostPlayerMiss

rodney528:


r6915ee:
What would be the primary use case for this callback?

rodney528:
> What would be the primary use case for this callback?

I thought of this because I made some code where I did `if (event.canceled) return;` in a script, but when doing `event.canceled = true;` somewhere else it didn't work for some reason.

Raltyro:
Honestly forgot to add this too aside with onPostPlayerHit or something

PR #745, opened 2025-08-16
Ctrl+Shift+F1 to force main menu state switch

superpowers04:


r6915ee:
What would be the use case of this in particular?

superpowers04:
If a state locks you or something, you can press Ctrl+Shift+F1 to force yourself back to the main menu

r6915ee:
Then I guess that can work, maybe. But I'd generally expect a more handy "restart mod" function if I say so myself. Regardless, I guess that makes sense.

NexIsDumb:
i think id rather have this in dev mode only, wdyt?
also add it in musicbeatsubstate atp

superpowers04:
> i think id rather have this in dev mode only, wdyt?

In my opinion, I think it'd be beneficial to always have this available as an emergency key combo to press if something gets stuck 

Especially since as far as I can tell, only the Windows build has a proper console

superpowers04:
> also add it in musicbeatsubstate atp

It runs regardless of if the substate disables persistantUpdates or not. So the keybind will work even if there's a substate

PR #748, opened 2025-08-17
Don't use resetSprite as a static extension

KarimAkra:
For some reason when using the static extension it would potentially cause a Null Object Reference when `resetSprite` calls `FlxTween.cancelTweensOf`
This fixes that issue (tho it makes no sense, it does lol)


NexIsDumb:
wtf
is this some kind of bug in either haxe or flixel
cuz im guessing its due to the recycle??

Raltyro:
Forgot what was i self assigned this for

NexIsDumb:
merging this just in case

PR #757, opened 2025-08-27
fixed character editor anim bug that happened when adding a new anim

KolzeYT:


NexIsDumb:
you showed me in dev, its good

PR #759, opened 2025-09-02
Fix MOD_ICON's misleading config comments

hifishlol:
The default config file generated for config.ini tells the user to add the .png extension when its already done, since it uses Paths.image

### DIFF
--- a/source/funkin/editors/ModConfigWarning.hx
+++ b/source/funkin/editors/ModConfigWarning.hx
@@ -22,10 +22,11 @@ API_VERSION=1
 DOWNLOAD_LINK="YOUR MOD PAGE LINK HERE"
 
 # Not supported yet
-;MOD_ICON64="path/to/icon64.png"
-;MOD_ICON32="path/to/icon32.png"
-;MOD_ICON16="path/to/icon16.png"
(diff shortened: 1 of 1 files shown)


source/funkin/editors/ModConfigWarning.hx
```diff
@@ -22,10 +22,11 @@ API_VERSION=1
...
 
 # Not supported yet
-;MOD_ICON64="path/to/icon64.png"
-;MOD_ICON32="path/to/icon32.png"
-;MOD_ICON16="path/to/icon16.png"
-ICON="path/to/icon.png"
+;MOD_ICON64="path/to/icon64"
+;MOD_ICON32="path/to/icon32"
+;MOD_ICON16="path/to/icon16"
+# The path starts in "Your Mod/Images/", do not add an image extension, it already defaults to png.
```
r6915ee, line 28: This might just be me being somewhat nitpick-y, but wouldn't it be better to put the folders as lowercase?

As in, something like `your-mod/images/`.

NexIsDumb:
honestly i think i could do something about this for detecting automatically extensions in the nearby future

PR #761, opened 2025-09-04
Fixed / Refactored SysZip to actually work, and store compressed entries without uncompressing an entire zip

ItsLJcool:
Refactored `SysZip` so that now it properly reads each `Entry` and saves only required information to unzip any `Entry` in the zip.
It no longer extends `Reader` since it's a custom implementation of unzipping by yours truely.

~~This commit also added a `CoolUtil` function for using the `haxe.Timer.stamp()` function to quickly see how long it takes for code to run.
Calling it once will start the timer, and call it again to finish the debug output. It also does fancy tracing. 👀~~

Edit:
This also adds a "cnemod" folder, this is how it can be loaded:
`./mods/ModFolder/cnemod.zip` or instead of `.zip` it can be a valid extension in `ALLOWED_ZIP_EXTENSIONS`


ItsLJcool:
There is an issue with streamed audio it seems like

ItsLJcool:
Seems the streamed audio bug was fixed, though now I found an issue with videos not loading properly so debugging that

ItsLJcool:
stupid fuckin hxvlc and VideoCutscene implementation

ItsLJcool:
there is still an issue with how videos are loaded so need to find a way to fix it...

ItsLJcool:
just fixed it by implementing what VideoCutscene did because it actually works amazing lmao

ItsLJcool:
After some testing it seems all the issues have been fixed.

The only issue that can occur is if the Mod owner doesn't change their `Flags.VIDEO_EXT` when trying to load other video extensions, causing it to not be detected in `ZipFolderLibrary` but easy to figure out probably?

ItsLJcool:
ok adding a feature in rq that pre-caches each video in the mod because unzipping loading time takes a hot sec, and videos easily get desynced

ItsLJcool:
Alr makred as ready now !!

unless I find another bug again lmao

ItsLJcool:
ughghghghghghhhhhh unready now, someone test zip mods with addons, and the new Flag I added to support custom zip extensions

ItsLJcool:
from my quick testing it seems like it works

ItsLJcool:
Ok this branch is ready to merge. The Streamed Audio bug just seems to be a mod specific issue, and happens rarely often on that mod.

NexIsDumb:
is this 100% done?

ItsLJcool:
> is this 100% done?

Doing final checks and it made me realize a flaw in how I wanted to load the `.cnemod` but fixing it right now

ItsLJcool:
Since I guess im stupid and can't realize it LMAO uh, basically the `.cnemod` isn't a real thing anyomre. Instead if you have a compressed library named `cnemod.zip` or a is in the `ALLOWED_ZIP_EXTENSIONS` then it will load that instead of the folder.

Solves a BUNCH of issues I was having, and will actually not fucking die on other OS's LMAO

Now everything is done and fix, that was the last lose end.

ItsLJcool:
ughhh need to fix literally every function that generates a script and search libraries to check instead of relying on openFL's Assets `exists` function

because compressed Libraries aren't folders, and are encrypted. So `Assets.exists` goes "huh not a folder, not valid!!!"

ItsLJcool:
NVM I was loading the mod wrong LMAO

NexIsDumb:
im just curious but if not even videos wont be uncompressed, i dont think the .temp folder would be needed anymore maybe?

NexIsDumb:
nvm, you're still saving the video

ItsLJcool:
> im just curious but if not even videos wont be uncompressed, i dont think the .temp folder would be needed anymore maybe?

Well no because of like Haxe limitation and just to be safe, we are extracting videos into .temp so it can decode the MP4 directly instead of unziping the video everytime we want to play it, which
1. Increases Lag
2. More time to decode
3. Storing the entire video in RAM

So it's just more Haxe friendly :)

ItsLJcool:
I will keep merging the branches cuz I am currently using this fork for a mod rn LMAO

Raltyro:
I was about to merge this since zips are basically fucked in cne but does videos works at all

ItsLJcool:
> I was about to merge this since zips are basically fucked in cne but does videos works at all

Yes this was a major issue I had to find an issue to fix lmao.

There was a class that depended on like the asset path returning `[ZIP]` and the path directly after, but with this commit that was removed and actually plays video cutscenes properly without having to check if the path contains a string, and will just use the method for getting the bytes directly from the ZipFolderLibrary, meaning it will check `.temp/` for the video, and also has saftey checks to remove any video files with the same name but not in the same folder.

It also precaches all the videos in your `.temp/` folder automatically but there really isn't a way to toggle it off yet, but there is code that supports it.

ZipFolderLibrary will only load videos based off of your `Flags.VIDEO_EXT` so if you don't change it in your mod / ini it won't properly load the video, but this only affects mods that don't care to think about the `Flags` class at all, when they really should and not fuck with things that might brick their mod.

[You can read this on why it was done this way](https://github.com/CodenameCrew/CodenameEngine/pull/761#issuecomment-3316146245)

Everything is really good to go, im still updating the repo and testing the Zip folder shit on a Pillar Funkin' dev mod build, but I also occasinally test zipping other mods to ensure nothing goes wrong.

ItsLJcool:
ts never being merged 
lj sadge

ItsLJcool:
Been testing with new mods, Bergabash, Dustin', SAU (from Zoro) works fine, there is a slight issue though??

`MultiFramesCollection` or something along the lines gave an error, then crashed when playing some Dustin' songs when I switched mods..? I think it might actually be a Dustin' issue cuz it doesn't occur in the normal build itself, though not too sure.

Raltyro:
MultiFramesCollection crashes happens in normal cne too though only when switching mods, also sorry lj for not responding

MRK2009:
IT GOT CONFLICTED, BRUH!!!

ItsLJcool:
So I think this is Merge worthy now, the only issues with this PR is the ones we can't solve without having dedicated workflows for Compressed Libraries.

Issues:
1. Accessing items in a directory in the Zip file requires you to decompress the folder and get the contents of the folder (each item inside is compressed)
2. Creating new files is impossible with Compressed Libraries inside the mod, so things like Autosaving, creating new files, modifying files, etc. cannot occur in Compressed Libraries.
3. Loading the Compressed Library for the first time takes ~1s per 1GiB (non-conclusive)
4. Anything related to `FileSystem`, `File`, etc won't work inside Compressed Libraries.

So we should advise mods that want to use Compressed Libraries to either account for these issues, or just don't use it.
This will be useful for addons, or big mods that would benefit from being compressed.

I tested out some mods but the results aren't really helpful, besides it verifying it properly decompresses files and doesn't bloat ram.

Since most mods didn't consider it being in a read-only state it's hard to gauge what is breaking and what is expected to be broken.

ItsLJcool:
Also it will take longer if the mod has a lot of videos to decompress, it will auto-decompress every video in the `./videos/` folder because it takes longer to decompress in runtime

MRK2009:
Okay now it's not conflicted, phew.

Raltyro:
is this good to merge?

ItsLJcool:
> is this good to merge?

Let me do one more passthrough and ill ping on discord when it's done, just making sure I didn't forget some stuff lol

ItsLJcool:
Ok @Raltyro its "good to merge" in the sense that if characters / animation offsets don't occur for other people it is good to merge, fsr I just got a random ass bug that doesn't properly load character offsets and this ik has occurred due to my external drive being shit, but I can't be 100% sure its a me issue

ItsLJcool:
w we are good to merge fixed all shit on my end so this all works now

PR #762, opened 2025-09-05
Add default font customization flags for FunkinText

r6915ee:
This pull request adds two new flags, that being `DEFAULT_FONT` and `DEFAULT_FONT_SIZE`. `DEFAULT_FONT` replaces the prior use of `vcr.tff` when loading the font in `FunkinText` but defaults to `vcr.tff`, while the size parameter in FunkinText's constructor has been changed to be purely optional without a default value, where having that parameter be null makes it so that it fetches `DEFAULT_FONT_SIZE` by default.

These changes intend to simplify the usage of custom fonts, especially in regards to `PlayState`.


NexIsDumb:
i dont see why not

PR #763, opened 2025-09-05
[LANG] pt-br

JDanielRandomizer:
BRAZIL COME TO BRAZIL UPDATEETETETGETETET


asmadeuxs:
I don't think it's necessary for us to have a Brazilian Portuguese translation because we can understand European Portuguese just fine, they just have minor vocabulary differences and different grammar that we STILL understand regardless
they say "Ficheiros" instead of "Arquivos" and we still understand the gist of it.

NexIsDumb:
agreed
ty tho!!

PR #771, opened 2025-09-17
Refactor Nix shell configuration for new Codename setup

r6915ee:
Codename has upgraded to Haxe 4.3.7 as the minimum recommended version, and the repository has moved some files around. The Nix shell configuration has been updated as a result.

The changes provided in this pull request include removing the version lock, and changing how the library-updating shell script is run.


r6915ee:
Do note that the library update shell script is currently not run as part of the shell script. The reason for this is because of the same issue with directories that #769 attempts to solve, where the directory appears to be malformed when running as part of the Nix shell hook that is run on shell startup, although the two branches have nothing to do with each other.

Possible solutions for that problem on Nix's side could include running the command line program directly, or executing the script like a program.

Additionally, the Haxe version provided is outdated. The reason for this is that Nixpkgs does not provide Haxe 4.3.7 at the moment, despite the version releasing in May of this year. An overlay is present in the code but in a commented state, due to compilation errors on the OCaml side of Haxe. Patches may be necessary, although I don't think that I can provide these patches (I have never used OCaml before).

These two factors are why this pull request, as of writing, is a draft pull request.

r6915ee:
The mentioned pull request is merged; I'll see if I can update the shell hook to use the newer shell script now.

r6915ee:
The shell script issue has been resolved now. It's just either setting up an overlay properly, using a community overlay, or waiting until Nixpkgs has the package.

r6915ee:
Apparently, `mbedtls` in the Haxe package is pinned at the unmaintained `mbedtls_2` package in the Nixpkgs repository instead of the standard version 3, with version 2 failing at some tests and causing the derivation build to fail; thus, the shell is not accessible. Because of this, in terms of pull requests, this one will be my highest priority until the pull request is finished (not exactly merged) or Haxe is updated over at Nixpkgs, since my current setup is reliant on the shell.

r6915ee:
I've pushed a small change that should hopefully counter the `mbedtls` issue; however, this pull request still isn't finished, as it still doesn't use Haxe 4.3.7.

r6915ee:
I've recently been switching my NixOS configuration to use a [Nix Flake](https://nixos.wiki/wiki/flakes) for simplification reasons, especially in regards to [Noctalia Shell](https://github.com/noctalia-dev/noctalia-shell), and, surprisingly, because of the Wine build failures over at NixOS Unstable's Hydra instance as a result of Nixpkgs upgrading Wine to version 11 and some major overhauls to the structure of the Wine packages. In that case, I'll likely also switch this shell configuration to use a Nix Flake instead, though that may potentially be a separate pull request.

r6915ee:
For now, I'm going to close this pull request. That's because I want to make a new one that makes use of a Flake instead. `flake-compat` will likely still be available to make use of the Flake in a non-Flake context.

PR #774, opened 2025-09-21
Add Opus audio playback support

HEIHUAa:
### Added OPUS audio playback support.
Also started adding multi-format extension support in ChartMetaData (though this part is unfinished). If you want to use OPUS audio globally right now, you can simply rename **.opus** files to **.ogg** — yes, they will still play correctly this way.

Also, OGG is itself a container format that supports Opus encoding. So you can absolutely encode your audio as Opus and package it within an OGG container.

### DIFF
--- a/source/funkin/backend/FunkinSprite.hx
+++ b/source/funkin/backend/FunkinSprite.hx
@@ -126,7 +126,7 @@ class FunkinSprite extends FlxSkewedSprite implements IBeatReceiver implements I
 		// hate how it looks like but hey at least its optimized and fast  - Nex
 		if (!debugMode && isAnimFinished()) {
 			var name = getAnimName() + '-loop';
-			if (hasAnimation(name))
+			if (hasAnim(name))
 				playAnim(name, null, lastAnimContext);
 		}
 	}
@@ -307,7 +307,7 @@ class FunkinSprite extends FlxSkewedSprite implements IBeatReceiver implements I
 
 	public function playAnim(AnimName:String, ?Force:Null<Bool>, Context:PlayAnimContext = NONE, Reversed:Bool = false, Frame:Int = 0):Void
 	{
-		if (AnimName == null)
+		if (AnimName == null || (!hasAnim(AnimName) && !debugMode))
 			return;
 
 		if (Force == null) {
@@ -318,16 +318,10 @@ class FunkinSprite extends FlxSkewedSprite implements IBeatReceiver implements I
 		if (animateAtlas != null)
 		{
 			@:privateAccess
-			// if (!animateAtlas.anim.animsMap.exists(AnimName) && !animateAtlas.anim.symbolDictionary.exists(AnimName)) return;
 			animateAtlas.anim.play(AnimName, Force, Reversed, Frame);
 			atlasPlayingAnim = AnimName;
 		}
-		else
-		{
-			if (!animation.exists(AnimName) && !debugMode)
-				return;
-			animation.play(AnimName, Force, Reversed, Frame);
-		}
+		else animation.play(AnimName, Force, Reversed, Frame);
 
 		var daOffset = getAnimOffset(AnimName);
 		frameOffset.set(daOffset.x, daOffset.y);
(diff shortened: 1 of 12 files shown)


HEIHUAa:
Also, I nearly forgot — **OGG is itself a container format that supports Opus encoding. So you can absolutely encode your audio as Opus and package it within an OGG container.**

NexIsDumb:
woah, i'll surely check this out

HEIHUAa:
> 哇，我肯定会检查这个的

What a coincidence! I just finished implementing the auto-detection feature.

Raltyro:
I was actually thinking of implementing it as a decoder so the getSound Assets function isnt a mess

HEIHUAa:
> 我实际上想将其实现为一个解码器，这样 getSound Assets 函数就不会一团糟了

cool

Vortex2Oblivion:
wowzers

HEIHUAa:
I'm wondering if **Opus** support can be implemented in **version 1.0.2**? Even if file extension recognition isn't added, that's acceptable.

If we assume no extension support is needed, then only the following three files from this pull request would require updates:

- building/libs.xml
- project.xml
- source/openfl/utils/Assets.hx

This approach would still allow using **Opus** audio encoding within the **.ogg** container.

Raltyro:
> I was actually thinking of implementing it as a decoder so the getSound Assets function isnt a mess

quoting my reply, i might be closing this pull request so dont do anymore to the pull request
i was thinking of implementing to the libraries in the organization as a decoder so it'll be easier to implement other audio types like MiniMP3 or Drmp3 for mp3, flac, etc.

HEIHUAa:
> > I was actually thinking of implementing it as a decoder so the getSound Assets function isnt a mess
> 
> quoting my reply, i might be closing this pull request so dont do anymore to the pull request i was thinking of implementing to the libraries in the organization as a decoder so it'll be easier to implement other audio types like MiniMP3 or Drmp3 for mp3, flac, etc.

Okay.

source/funkin/backend/system/Flags.hx
```diff
@@ -56,6 +56,11 @@ class Flags {
 	public static var REPO_OWNER:String = "CodenameCrew";
 	public static var REPO_URL:String = 'https://github.com/$REPO_OWNER/$REPO_NAME';
 
+	public static var SOUND_EXTENSIONS:Array<String> = [
+		#if web "mp3" #else "ogg" #end,
+		"opus"
```
r6915ee: Wouldn't it make more sense to include this as part of the `#else` macro, since Opus support is only available on C++ builds?

HEIHUAa:
I was thinking, since CNE has hxvlc installed, could we use VLC to decode audio? In theory, this would allow CNE to support a wide range of audio formats — though I’m not sure how to actually implement it.

r6915ee:
> I was thinking, since CNE has hxvlc installed, could we use VLC to decode audio? In theory, this would allow CNE to support a wide range of audio formats — though I’m not sure how to actually implement it.

I'm not familiar with how `hxvlc` properly interfaces with the actual C library, but `hxvlc` is primarily intended for video playback. `hxvlc` does not provide any features for audio management with the exception of manually managing `libvlc` through its respective class. This factor would also make Codename more dependent on desktop targets and prevent some compatibility with tools like Hashlink. `hxvlc` is also only usable on desktop targets, because `libvlc` is provided only on those targets; web targets would still need to use their own system.

HEIHUAa:
> > 
> 
> I'm not familiar with how `hxvlc` properly interfaces with the actual C library, but `hxvlc` is primarily intended for video playback. `hxvlc` does not provide any features for audio management with the exception of manually managing `libvlc` through its respective class. This factor would also make Codename more dependent on desktop targets and prevent some compatibility with tools like Hashlink. `hxvlc` is also only usable on desktop targets, because `libvlc` is provided only on those targets; web targets would still need to use their own system.

Okay, I understand. Thank you for explaining.

Raltyro:
Opus is now implemented in internal-merge

HEIHUAa:
> Opus is now implemented in internal-merge

cool

PR #780, opened 2025-09-23
KeybindsOptions: Default categories accessible outside of the state, translation support for custom options

Bitto1070:
- Moved the categories outside the "create" function to let them become accessible to other states when KeybindsOptions is imported;
    - Theoretically, this shouldn't cause any issues, as it's not replacing the current categories variable.
- Added translation support to custom options;
    -  It checks if the translation even exists first before translating, to not cause any issues with old options.xml files (hopefully).
- custom options menus are now tagged with a "custom: true", like their settings;
    - Really only used to check if it should add the "KeybindsOptions." prefix before translating. Might be useful to someone else, though.
        - oh yeah i also kinda removed that prefix from the translate function so people aren't forced to use it on their options.xml if they want to add translation support
- also makes isSubstate public

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -34,3 +34,7 @@ f8b20774e3f035eeddb9211a28a391b20db74456
 # Revert a revert (empty script)
 392416e00cbd82e1c49e82a9fafdcf2e069a654e
 159467de832a172681200c463d244a46e295c3c8
+
+# Revert a revert (scripted cutscene path)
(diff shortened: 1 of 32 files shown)


HeroEyad:
this is very good, i like this.

HeroEyad:
@NexIsDumb thoughts?

NexIsDumb:
oops
sorry, imma check it tmr promised

NexIsDumb:
alrr, checked, looks fine to me

PR #783, opened 2025-09-27
Add variation to Week.hx so it works in storymode

Mooshie318:
variations don't work in story mode even though it's a variable in the xml, so I made it work


Raltyro:
Huh, mustve been the wind - me when adding variations

MRK2009:
Why is this not merged?😭

Raltyro:
i forgot

MRK2009:
> i forgot

Ty
Finally

PR #785, opened 2025-09-30
push event to event list only if it doesnt exists yet

lasystuff:


Raltyro:
DId you forget to wrap the whole code thats responsible for reading the event data?

lasystuff:
oops sorry

PR #790, opened 2025-10-11
Update macOS version to 14.

MRK2009:
Reason: macOS 13 will be removed on 4th December in Github.


MRK2009:
TYSM RALTY!!!

Raltyro:
Should've tested it first in your repository
It's not compiling for MacOS

PR #792, opened 2025-10-15
I HATE CAMZOOMMULTI

chronicsilly:
image

frakits approved ok

\>change the way camera bopping works
\>don't even intergrate it into the engine properly
\>break zoomFactor
w codename engine


Raltyro:
zoomFactor not being accounted in camera.zoomMultiplier is intended, though this is a good change too

Raltyro:
Maybe change as a default to account zoomMultiplier in the future? no lcue ill probably forget this and its all my fault lol...

chronicsilly:
> zoomFactor not being accounted in camera.zoomMultiplier is intended, though this is a good change too

...why? And if so why not make it disabled by default so people don't assume the engine is broken

PR #793, opened 2025-10-15
Charter/Editor additions and fixes

TheZoroForce240:
List of changes/fixes

- Camera highlights in charter grid
- Reworked UIContextMenu to support submenus and sliders
- Moved Waveforms, Note Types and Bookmarks into submenus
- Added volume sliders for audio related options (Replaced old mute options)
- Added global hitsound volume slider
- Strumline vocals mute/volume option no longer shows if it has no vocals loaded
- Strumline preview in the edit menu now is forced to the default window size with an area outline when resizable editors is enabled, rather than being scaled incorrectly
- Fixed the Editor SFX debug option and now should actually disable all editor sounds if turned off
- Fixed note type switching with keybinds (never worked before)
- Fixed saving difficulty specific chart metadata

### DIFF
--- a/source/funkin/editors/charter/Charter.hx
+++ b/source/funkin/editors/charter/Charter.hx
@@ -81,7 +81,6 @@ class Charter extends UIState {
 	public var metronome:FlxSound;
 
 	public var vocals:FlxSound;
-	public var voicesMuted:Bool = false;
 
 	public var quant:Int = 16;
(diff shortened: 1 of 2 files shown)


MRK2009:
Bruh, it conflicted.

MRK2009:
Phew it's no longer conflicted.

PR #805, opened 2025-10-30
Moving the PauseSubstate background variable declaration out of the create function

SorbetLover:
This will add the possibility to reference it on the code


FuroYT:
I was planning on using `FlxCamera::bgColor` (like how VSlice does it) instead of making a sprite for a bg

SorbetLover:
Would be cool, but meanwhile you dont make this change id like to be able to reference the bg on the code

PR #813, opened 2025-11-06
Better mouse overlapping function for CoolUtil

Bitto1070:
its FlxG.mouse.overlaps, but working on objects in cameras that arent FlxG.camera at scroll 0
probably needs a better description but thats the best i could come up with


ItsLJcool:
> its FlxG.mouse.overlaps, but working on objects in cameras that arent FlxG.camera at scroll 0 probably needs a better description but thats the best i could come up with

Here is a better description:

Implements a CoolUtil function that allows you to properly check if the Mouse overlaps an object on any camera layer, properly.

source/funkin/backend/utils/CoolUtil.hx
```diff
@@ -1003,6 +1003,21 @@ final class CoolUtil
...
+	/**
+	 * Returns if the mouse is overlapping the sprite on the given camera, taking the camera's position and scroll into account.
+	 *
+	 * @param   sprite  Any `FlxBasic`
+	 * @param   camera  The desired "screen" to check overlap on. If `null`, the first camera in `FlxG.state.cameras` is used (normally `FlxG.camera`).
+	 * @return  Bool
+	 */
+	public static function mouseOverlaps(sprite:FlxBasic, ?camera:FlxCamera) {
+		if (camera == null)
+			camera = FlxG.state.cameras[0];
```
ItsLJcool: You should change this to be `camera = FlxG.camera;` because that's the default camera. this could possibly be null by itself

Bitto1070:
i meant for the actual function in the code but thanks lj

chronicsilly:
you should define the type of posthing as FlxPoint and also wouldn't it be better if it used the sprite's camera by default rather than the first camera in the list?

Bitto1070:
> wouldn't it be better if it used the sprite's camera by default rather than the first camera in the list?

somehow i never thought about doing this, let me test and push it

chronicsilly:
> > wouldn't it be better if it used the sprite's camera by default rather than the first camera in the list?
> 
> somehow i never thought about doing this, let me test and push it

Reason why I bought it up is I have similar code running on some of my mods as a public function in global.hx that check for the sprites camera

Having it in coolutil or just as FlxG.mouse.overlaps would be a huge help too

NexIsDumb:
honestly i fw this

PR #814, opened 2025-11-07
Improve documentation generation scripts

r6915ee:
This pull request improves the automation of documentation generation.

From the initial commit's message:

> There is now a UNIX documentation generation script in the form of a shell script. The Windows shell script has also been improved, and some of the output of that script has been improved to match the UNIX one.
>
> Because the scripts still use Lime to generate the documentation, the shell script checks the kernel name using `uname`, and changes the documentation generation command ever so slightly. The reason for this is that Lime only supports building to one platform at a time. This makes manual intervention necessary, so the UNIX script checks for the kernel ahead of time.

The filenames of the scripts now follow snake-case and follow a similar naming convention to the library installation scripts. Do note that Lime is also invoked through Haxelib for compatibility reasons (particularly with NixOS), but it should work the same and maybe even somewhat faster.

### DIFF
--- a/building/README.md
+++ b/building/README.md
@@ -1,6 +1,6 @@
-Here's the full guide on how to setup and compile Codename Engine!<br>
-
-> **Open the instructions for your platform**
+# Compiling Codename Engine
+Do you want to turn your source code into a playable build to play? Then you want to **compile the source code**, follow this guide.
+> **Open the instructions for your platform.**
 <details>
     <summary>Windows</summary>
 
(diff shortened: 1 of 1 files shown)


r6915ee:
The UNIX documentation generation script has just been updated to only check for the Darwin kernel used in macOS, and assume every other platform as Linux, considering Linux-compatible platforms such as FreeBSD and custom kernels.

Raltyro:
I'm not too knowledgable to even touch the documentation scripts, so i'm requesting a review who does

Raltyro:
Please anyone?

MRK2009:
3 at the same time lol.

MRK2009:
@NexIsDumb sorry for the disturb, do you know anything about the documentation script or no?

NexIsDumb:
this looks fine to me as its basically just a slightly modified copy paste of the older version, way better now though
but i was thinking, should we move these to /building maybe, wdyt?

NexIsDumb:
yeah okay it very should, gimme a sec

r6915ee:
> this looks fine to me as its basically just a slightly modified copy paste of the older version, way better now though but i was thinking, should we move these to /building maybe, wdyt?

I'm fine with that; it generally makes more sense than the `/art` directory.

NexIsDumb:
added a docs generation guide too

PR #789, opened 2025-10-10
Add onCloseSubState to MusicBeatState so that scripts can know when a SubState closes, and can even prevnt it

ItsLJcool:
this isn't tested btw i just made a change form google LMAO

PR #817, opened 2025-11-14
Add `onCloseSubState` script event to `MusicBeatSubstate`

richTrash21:
Adds `onCloseSubState` script event for `MusicBeatSubstate`s. Basicaly #789 but for substates lol

### DIFF
--- a/source/funkin/backend/MusicBeatSubstate.hx
+++ b/source/funkin/backend/MusicBeatSubstate.hx
@@ -236,6 +236,12 @@ class MusicBeatSubstate extends FlxSubState implements IBeatCancellableReceiver
 			super.openSubState(e.substate is FlxSubState ? cast e.substate : subState);
 	}
 
+	public override function closeSubState() {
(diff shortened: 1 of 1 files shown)


ItsLJcool:
yes this would be nice... if only there was an `onClose` event callback because this is already implemented :trollface:
https://github.com/CodenameCrew/CodenameEngine/blob/main/source/funkin/backend/MusicBeatSubstate.hx#L161

ItsLJcool:
the only reason I added it into `MusicBeatState` because there was no physical event that was called when the SubState closed. This is not true for the `MusicBeatSubState`. This does prove useful however for substate substates, so I still think it's a valid pr

richTrash21:
> This does prove useful however for substate substates

Thats why I made this pr! I'm using `onOpenSubState`/`onCloseSubState` to turn on/off mouse events on menu buttons and one of the menus is a substate... that can open other substate menu. 😅😅

PR #818, opened 2025-11-15
Support for `hxvlc 2.0.0` and higher

richTrash21:
Changes:
- `VideoCutscene`: Fixed compiling with `hxvlc 2.0.0` and higher.
- ~libs.xml: hxvlc 1.9.3 => hxvlc 2.2.4.~ Reverted to not break mods.

### DIFF
--- a/building/libs.xml
+++ b/building/libs.xml
@@ -17,7 +17,7 @@
 		<git name="flxanimate" url="https://github.com/CodenameCrew/cne-flxanimate" />
 		<git name="hxdiscord_rpc" url="https://github.com/CodenameCrew/cne-hxdiscord_rpc" skipDeps="true" />
 		<lib name="funkin-modchart" version="1.2.4" skipDeps="true" />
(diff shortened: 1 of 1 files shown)


ItsLJcool:
this will break some mods iirc since I think it changes how you access some variables, as FlxVideoSprite no longer uses`FlxVideo` directly but uses `FlxInternalVideo` which will 100% break my implementation of Caching videos.

This should be merged before v1.0.2 imo

ItsLJcool:
Here is the current commit [CodenameEngine](https://github.com/MAJigsaw77/hxvlc/blob/88582c6c72adb3f00c450de3c8ff64b9e3307d76/source/hxvlc/flixel/FlxVideoSprite.hx) uses as of rn (I had it saved cuz it pissed me off so much)

richTrash21:
> this will break some mods iirc

Yeah right, my bad. 😅😅

Raltyro:
this is good for anyone who just wants to compile cne with newer haxelibs

PR #820, opened 2025-11-16
Minor FPS Optimizations

HEIHUAa:
Limited FPS counter updates to 15 times per second
- Reduces performance and memory impact caused by frequent text redrawing
- The FPS counter will no longer flicker uncontrollably, making it finally readable.

Replaced FlxG.rawElapsed with native __enterFrame delta time
- Maintains accurate FPS display when window loses focus (assuming game pauses on focus loss)
- Removes the minimum 10 FPS display limit

### DIFF
--- a/source/funkin/backend/system/framerate/SystemInfo.hx
+++ b/source/funkin/backend/system/framerate/SystemInfo.hx
@@ -154,20 +154,13 @@ class SystemInfo extends FramerateCategory {
 		if (totalMem != "Unknown" && memType != "Unknown") __formattedSysText += '\nTotal MEM: $totalMem $memType';
 	}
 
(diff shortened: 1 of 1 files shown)


Raltyro:
Don't use __enterFrame deltatime, as it's inconsistent and varies in openfl versions

HEIHUAa:
> 不要使用 __enterFrame deltatime，因为它不一致，并且在 openfl 版本中变化

I've switched the delta time calculation to use FlxG.rawElapsed + openfl.Lib.getTimer(), and changed the frame rate display to use rounding.

HEIHUAa:
Fixed the issue where VRAM display showed -2147483648B.

HEIHUAa:
@Raltyro I think the modifications should be done now. Could you take a look?

PR #833, opened 2025-11-21
Improve Accurate VocalsResync

HEIHUAa:
Updates to the synchronization detection based on the newly modified audio.


Raltyro:
image
Maybe 4ms threshold is too strict? can we tune it to 10ms or 7ms instead?

HEIHUAa:
image
image
Even with detection running every frame, I conducted some stress tests. Under these conditions, it only occasionally shows a 4-6ms spike on a single frame. Since there are still untested scenarios, I've decided to stay conservative for now.

HEIHUAa:
Modified to 6ms.

PR #835, opened 2025-11-21
Quick fix allowing hits to be counted even if showRating is false

Jurtaa:
I overlooked a specific detail in one of my last pull requests, so I'm making this one to rectify that. This should hopefully allow `hits` to be tracked even if `event.showRating` is set to false. JS Engine


Raltyro:
what is JS Engine

r6915ee:
> what is JS Engine

[JordanSantiagoYT/FNF-JS-Engine](https://github.com/JordanSantiagoYT/FNF-JS-Engine).

PR #838, opened 2025-11-26
Update macOS runner version to 15-intel

MRK2009:
This is the only intel-based of macOS is free, since macOS 14 large is only for money, even though 15 is kinda slow, but works anyways.
Btw it's only available until 2027.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     name: Mac OS Build
     permissions: write-all
-    runs-on: macos-13
+    runs-on: macos-15-intel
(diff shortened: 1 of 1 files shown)


MRK2009:
Somehow it got cancelled, weird.

Raltyro:
Oh what

MRK2009:
> Oh what

I was trying to use Intel based of macos 15 since it doesn't use arm architecture, and it looks its working, but somehow it got cancelled for being too long, idk why.

MRK2009:
Ig the only way to fix, is the cne-lime fork, to have arm architecture support for regular macos 14/15.

PR #840, opened 2025-11-29
Don't pause persistent sounds in `FunkinParentDisabler`

richTrash21:
Self explanatory. Also spelling mistake in commit title oops😅😅


richTrash21:
wow that was fast

PR #851, opened 2025-12-14
Minor credits fix

MRK2009:
Since Theo changed his username, I had to fix the link, so people won't get confused of his account got deleted or not. Also I added links for both MAJigsaw and Smokey, so they can immediately go to see their github accounts.


MRK2009:
If I'm doing anything wrong, review it.

Raltyro:
i'm, not going to say anything, though i will be merging the link fixes for majigsaw and smokey though

MRK2009:
Okay

PR #856, opened 2025-12-18
Add a way of "cancelling" updateSustainClip

sillyburg:
just as the title says, with this you can cancel the sustain's clipRect from updating in a NoteHitEvent using preventSustainClip.

i thought this was always a thing but apparently not from what i checked?

not the best but it'll work


sillyburg:
please codename... i Need this... my note is kinda missless......

sillyburg:
thank you codename... 🥹

PR #864, opened 2025-12-31
the fucking sfx delay that i hate

KolzeYT:
pretty self explanatory isnt it? the reason that i made this pr is that i did a on step scroll thing like in the character selector for vslice for one of the mods that im working on and the scroll sfx was delayed


karucodesstuff:
that fucking sfx delay that i hate

PR #868, opened 2026-01-11
FlxView3D (Away3D) Optimisation

kitzsh:
This pull request optimises Codename Engine's Away3D implementation by rendering the view directly to a framebuffer, rather than taking a snapshot of the output and loading it onto a BitmapData.
This eliminates a bottleneck which previously caused performance issues which got worse as the view's resolution increased.


Raltyro:
Is this complete?

kitzsh:
> Is this complete?

There's a bug which I used a hacky workaround to fix that might be good to look into further, but apart from that it's basically done

PR #874, opened 2026-01-18
Removing CoolUtil.quantize from note budgeting

HEIHUAa:
I don't think this is necessary and it might consume a small amount of performance. Attempting to use Tween instead could potentially lead to precision issues.


Raltyro:
your like the only person that are still able to contribute and i respect it

PR #882, opened 2026-01-21
Add ability to set custom alternate animation suffix

r6915ee:
This pull request simply adds the ability to set the alt. animation suffix. As part of this:

- A new flag was added, `DEFAULT_ALT_ANIM_SUFFIX`, that sets the default suffix for alt. animations. By default, it is set to `-alt` in order to retain compatibility.
- A new property for `StrumLine` was introduced, that being `defaultAnimSuffix`, which is used to replace the suffix per-strumline. By default, it is set to the `DEFAULT_ALT_ANIM_SUFFIX` flag mentioned above.
- `Alt Animation Toggle` will automatically make use of the property above.

This change is primarily intended to technically allow for much more alt. animations on a single character without going through the process of a complete character change, thus acting as a more performant variant of theoretically "switching" characters.

Do note that, although the `Alt Animation Toggle` event has been updated to take advantage of the new per-strumline property, it did not receive any changes to allow setting the suffix directly as that is outside of the scope of the pull request (at least, in my opinion).


r6915ee:
Now that I think about it, I'll probably add support for prefixes as well, as it makes sense like with the window titles. I'll put it back as a draft pull request for now and then convert it once I've made the change. (I'm not renaming the branch for this, I don't want to remake the pull request.)

r6915ee:
Okay, there are some issues with events in particular (recycling appears to be definition-dependent in terms of parameter location) that prevent the easy implementation of prefixes. As a result, I'll ready the pull request for review again in its current state, but in my case I'll keep the animation prefix feature in mind; I have some ideas on how to implement that, alongside a potential backward-incompatible rewrite to the character animation system, but those are outside the scope of this pull request at the moment.

PR #888, opened 2026-01-25
Add persist parameter to playMenuSFX function

NebulaStellaNova:
NO clue why this wasn't already a thing :D


Raltyro:
Rejecting this... since you can already do
```hx
var snd = CoolUtil.playMenuSFX(CANCEL, 1);
snd.persist = true;
```

PR #890, opened 2026-01-25
Enable 'VIDEO_CUTSCENES' for iOS

MRK2009:
Since it uses the latest version of hxvlc instead of 1.9.3, it definitely can work on it, so there's no point of leaving it as only Android, if it already got fixed for iOS


MRK2009:
I forgot to update it, oopsies.
Just ignore it

MRK2009:
Wait, I'm being stupid, why did I the branch, sorry about it, I kept rushing ahead.

PR #894, opened 2026-01-30
BloomEffect

HEIHUAa:
This bloom effect is modified based on OpenFL's BlurFilter, achieving a result that is difficult for HScript to replicate with equivalent performance. Currently in its early stage and merely functional, I anticipate adding more optimization settings, such as bloom blend modes, custom brightness weighting, among other visual options, as well as improving the comments.
PotPlayerMini64_f9La4nl6j0


nebulazorua:
Actual really cool usage of bitmapfilter actually

I just feel like instead of a specific bloomeffect though, might be better to make a generic MultiPassShaderFilter with a bloom filter as an example?

HEIHUAa:
> Actual really cool usage of bitmapfilter actually
> 
> I just feel like instead of a specific bloomeffect though, might be better to make a generic MultiPassShaderFilter with a bloom filter as an example?

Yes, that's exactly what I thought before working on this. Unfortunately, I'm not very skilled at writing source code, so I can't implement this feature myself.However, if this bloom effect meets my optimization goals, it should still deliver decent performance on most low-end devices. I'm thinking of something like the bloom quality you see in Roblox?

NexIsDumb:
oh my???

HEIHUAa:
**Currently supports blurX and blurY.**
**Currently, `quality` refers to the number of samples. However, since lowering this value further will cause noticeable ghosting issues, it will later be changed to represent the bloom rendering resolution.**
**`strength` is the intensity of the bloom effect.**
**`threshold` determines how bright each RGB component needs to be before bloom is triggered.**

**Future additions may include:**
- **Gradual threshold**: The closer the pixel color is to the threshold, the lower its influence on the bloom brightness.
- **Bloom brightness weighting**: A simple, user-definable brightness weighting to increase customization.
- **Blend mode**: The blending mode used between the bloom effect and the original image.

(If I can successfully implement a reduction in the bloom resolution, the performance will improve by leaps and bounds. Even if it's only rendering at 270p while outputting at 1080p, the visual result would still be perfectly acceptable. The number of pixels the bloom effect needs to process would then be just 6.25% of the original amount.)

HEIHUAa:
Current testing is being done on a mobile device.  
With `useLowQualityExtract` enabled, a bloom range of 100x100, `quality = 24`, and all other settings at default, it achieves around 50 FPS on a Dimensity 720. However, `useLowQualityExtract` is not well optimized on mobile—disabling it results in at least 60 FPS. I'm not sure what the maximum FPS could be, but I think the performance is sufficient.

HEIHUAa:
I think this should be done by now.

HEIHUAa:
@Raltyro I don't mean to bother you, but I'd like to use it. My mod needs this, and it doesn't seem to have any issues right now.

PR #895, opened 2026-01-30
Always use lastest FunkinModchart's release version

theoo-h:
To avoid having to do PR every time I update FM, I suggest simply using the latest and most stable version (currently 1.2.5), which fixes many bugs, including a crash that occurred every single time you exited a song.


NexIsDumb:
sure

PR #896, opened 2026-01-30
Allow string values for radio buttons in options menu :D

NebulaStellaNova:
NOT sure why this wasn't already a thing.... But it's here now


dotaxel (changes requested):
Iirc Std results can variate over platforms so i'd recommend using Math.isNaN instead of comparing like that

dotaxel:
@Raltyro comparing to Math.NaN its always false !!!!

Raltyro:
I forgot thank you

PR #897, opened 2026-02-01
Re-implement onDraw functionality for FunkinSprite

ItsLJcool:
As FlxAnimate didn't have the functionality before, so we have to do the check in FunkinSprite ourselves, as it itself stop's calling the `super.draw();` when it is a `isAnimate`


Raltyro:
LJ

ItsLJcool:
> LJ

Sorry I just made the fork and forgot about it

Also usually iffy about testing anything in cne cuz I usually break everything when I commit lol

PR #898, opened 2026-02-02
SUSTAINS_AS_ONE_NOTE = MOD_API_VERSION < 2

HEIHUAa:
**Shouldn't `SUSTAINS_AS_ONE_NOTE` be enabled when `MOD_API_VERSION = 1`?**
**According to my understanding, `MOD_API_VERSION = 2` should be considered the legacy API.**


Raltyro:
sustains as one note is a new thing in api version 2, it doesnt exists in cne 1.0.0 or 1.0.1 which is api version 1,
i was going to say thisll break compatibility but, alot of recent changes already breaks compatibility...

HEIHUAa:
> sustains as one note is a new thing in api version 2, it doesnt exists in cne 1.0.0 or 1.0.1 which is api version 1,
> i was going to say thisll break compatibility but, alot of recent changes already breaks compatibility...

Alright, I misunderstood. Sorry for the trouble.

PR #899, opened 2026-02-03
Fix NaN check in RadioButton options

dotaxel:
Comparing to NaN (even with itself, IEEE-754 standard rules) its always false, use Math.isNaN insted


FlameCookie123006:
I would like you to add an event to change the mania (that is, the amount of keys in the strum).

dotaxel:
no

FlameCookie123006:
> no

Hey, it's okay, at least I tried to contribute something.

PR #900, opened 2026-02-03
Improved text handling and keyboard input functionality for UITextBox

TerameTechYT:
My one main gripe with the editors was the text boxes, they were a bit of a hassle to use, so I decided to implement a few things I thought would improve the usability of them quite a bit.

- Clamp position between 0 and text length (no more wrapping around)
- Disable sounds on modifier keys (control, alt, shift)
- Delete/Backspace for whole words (control + delete/backspace)
- Cut (control + x)
- Use cut paste copy sounds


ItsLJcool:
w committer

PR #901, opened 2026-02-05
optimize __notePerStrum

HEIHUAa:
**The optimization ensures that the `Vector` is not recreated each time, preventing new memory allocation on every operation.**
**Additionally, the ghostTapping detection conditions have been merged.**
**Added an early exit to the `__inputProcessJustPressed` function.**


MRK2009:
Noice

PR #914, opened 2026-02-27
add "onPostEvent" call

sillyburg:
help me idk how to make pull requests

Por favor velocidad necesito esto mi madre es algo pobre


sillyburg:
idk why it added the prevent sustain clip commit again Help me

ItsLJcool:
the `preventSustainClip` is already in the current source code so it doesn't matter.

Also why do we need a post event function for events?

sillyburg:
> the `preventSustainClip` is already in the current source code so it doesn't matter.
> 
> Also why do we need a post event function for events?

wanted to modify a specific behaviour of Camera Movement, and since onEvent is called before the hard-coded events i have to delay the functions by one frame and i don't like doing that

ItsLJcool:
> > the `preventSustainClip` is already in the current source code so it doesn't matter.
> > Also why do we need a post event function for events?
> 
> wanted to modify a specific behaviour of Camera Movement, and since onEvent is called before the hard-coded events i have to delay the functions by one frame and i don't like doing that

You can cancel the event you know, right?
```haxe
function onEvent(e:EventGameEvent) {
    if (e.event.name != "Camera Movement") return; // so we only check for this event
    e.cancel();
    
    // write custom logic
}
```

sillyburg:
> > > the `preventSustainClip` is already in the current source code so it doesn't matter.
> > > Also why do we need a post event function for events?
> > 
> > 
> > wanted to modify a specific behaviour of Camera Movement, and since onEvent is called before the hard-coded events i have to delay the functions by one frame and i don't like doing that
> 
> You can cancel the event you know, right?
> 
> ```hx
> function onEvent(e:EventGameEvent) {
>     if (e.event.name != "Camera Movement") return; // so we only check for this event
>     e.cancel();
>     
>     // write custom logic
> }
> ```

i want to modify certain properties of the camera movement tween without having to rewrite the entire thing and without having to delay the function by one frame

PR #921, opened 2026-03-05
f11 fullscreen

HEIHUAa:
**Add F11 fullscreen functionality.**  
**Remove F key fullscreen functionality from `TitleState`.**


source/funkin/backend/system/FunkinGame.hx
```diff
@@ -1,9 +1,26 @@
...
 class FunkinGame extends FlxGame {
 	var skipNextTickUpdate:Bool = false;
+	
+	var fullscreenListener:KeyboardEvent->Void;
+	
+	public function new(gameWidth:Int, gameHeight:Int, entryState:Class<FlxState>, updateFramerate:Int = 60, drawFramerate:Int = 60, skipSplash:Bool = false, startFullscreen:Bool = false) {
+		super(gameWidth, gameHeight, entryState, updateFramerate, drawFramerate, skipSplash, startFullscreen);
+		
+		fullscreenListener = function(e:KeyboardEvent) {
+			if (e.keyCode == 122) {
```
ItsLJcool: You should not use magic numbers and use the FlxKey abstract instead

r6915ee:
Isn't this already provided through the Alt + Enter keybind?

HEIHUAa:
> Isn't this already provided through the Alt + Enter keybind?

It conflicts with the Enter confirmation key, causing unintended confirmation actions when I toggle fullscreen. Since I switch between fullscreen and windowed mode frequently, this is quite annoying.

MRK2009 (approved):
I agree with that, nice.

source/funkin/backend/system/FunkinGame.hx
```diff
@@ -1,9 +1,28 @@
...
 class FunkinGame extends FlxGame {
 	var skipNextTickUpdate:Bool = false;
+
+	#if desktop
+	var fullscreenListener:KeyboardEvent->Void;
+	
+	public function new(gameWidth:Int, gameHeight:Int, entryState:Class<FlxState>, updateFramerate:Int = 60, drawFramerate:Int = 60, skipSplash:Bool = false, startFullscreen:Bool = false) {
+		super(gameWidth, gameHeight, entryState, updateFramerate, drawFramerate, skipSplash, startFullscreen);
+		
+		fullscreenListener = function(e:KeyboardEvent) {
```
Raltyro: Can it be in `override create(_):Void` instead of in `public function new`?
and the fullscreenListener function should be just a normal function

Raltyro (changes requested):

source/funkin/backend/system/FunkinGame.hx
HEIHUAa: Makes sense, I'll change it later.

MRK2009:
You're almost finished.
One more thing and then done.

PR #924, opened 2026-03-19
String buf

HEIHUAa:
**Use `StringBuf` to reduce memory overhead from string concatenation.**


r6915ee (commented):
I'm noticing that the way the buffer is modified is done using large sections of code that can consist of up to ten lines using just the `add` method. I'd personally recommend iterating through every modification and applying it that way.

```haxe
var buf:StringBuf = new StringBuf();
for (i in ["prefix", data, "suffix"])
    buf.add(i);
```

HEIHUAa:
> ```hx
> for (i in ["prefix", data, "suffix"])
> ```

**Actually, in the `for (i in ["prefix", data, "suffix"])` operation, the `[]` creates a new array, allocating new memory.**

HEIHUAa:
I ended up writing a macro for this. It might make the code look a bit bloated, but it's definitely better than the previous approach. Of course, if there's a better way, feel free to let me know.

r6915ee (approved):

HEIHUAa:
**Use a macro to automatically convert `$` to `StringBuf`, preserving readability as much as possible.**

Raltyro:
Why is there a macro?

HEIHUAa:
> Why is there a macro?

Since I wasn't sure how to use `StringBuf` in the most concise and readable way, you can check out https://github.com/CodenameCrew/CodenameEngine/pull/924/changes/278eaa119d2860ec3ff6f87d5857cb296ed8290b to see that the previous approach basically made concatenated text nearly unreadable.

PR #933, opened 2026-03-28
Add Text displaying if the user is running an Experimental Version of CodenameEngine

ItsLJcool:
This will not effect releases as it uses a different workflow with different commands, so this will help users who need help or reminder of what Experimental Build they are using to help looking at the source code comparisons.

Image showcasing the new '(Experimental Build)' text when running an experimental build.


MRK2009:
Noice

PR #941, opened 2026-03-30
Optimize arrow handling

HEIHUAa:
First, move the `downscrollCam` position calculation before the `isOnScreen` check to prevent culling errors.

Simplify the bloated `downscrollCam` calculation:

```haxe
// Before:
final xx = x;
x += origin.x - offset.x;
x -= __strum.x; x *= -1; x += __strum.x;
x -= origin.x - offset.x;
x += __strum.width; // ??? maybe this isnt good

// After:
x = -x + 2 * (__strum.x - origin.x + offset.x) + __strum.width;
```
These are actually equivalent.

Additionally, implement caching for extensive arrow calculations. If the calculation for an arrow remains unchanged from the previous one, reuse the previous result. This significantly reduces computation. For example, `__noteAngle` previously required a separate calculation for each arrow. After this change, if all arrows in the list are the same, it will only be calculated once.

Remove `abs` from `scrollSpeed`. Since there's no need to ensure correctness during runtime, if users find errors related to this, they can correct them in their own large codebases. Some may even intentionally want that "incorrect" effect.

Remove the setter in the calculations within `updateSustainClip` and `set_clipRect`, keeping only the values that are actually needed from the original setter.

**Finally, removed the unused static variable `__notePosFrameOffset`. Changed the static variable `__posPoint` to a local variable, making the code cleaner and improving performance.**

### DIFF
--- a/source/funkin/game/Note.hx
+++ b/source/funkin/game/Note.hx
@@ -272,8 +272,8 @@ class Note extends FlxSprite
 			if (__noteAngle != __lastAngle) {
 				__lastAngle = __noteAngle;
 				final result = FlxMath.fastSinCos((__noteAngle + 90) * FlxAngle.TO_RAD);
-				__lastAngleCos = result.cos;
 				__lastAngleSin = result.sin;
+				__lastAngleCos = result.cos;
 			}
 
 			if (__strum.width != __lastStrumW || __strum.height != __lastStrumH) {
--- a/source/funkin/game/Note.hx
+++ b/source/funkin/game/Note.hx
@@ -240,17 +240,6 @@ class Note extends FlxSprite
 	 */
 	public var strumRelativePos:Bool = true;
 
-	override public function isOnScreen(?camera:FlxCamera):Bool {
-		var downscrollCam = (Std.isOfType(camera, HudCamera) ? cast(camera, HudCamera).downscroll : false);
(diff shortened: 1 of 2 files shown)


betpowo:
how did i not see any of this   . thank u for this btw....

HEIHUAa:
Another optimization was also added: the per-frame operation for `__cachedCopyFields` has been improved. A condition check is now added — if `__cachedCopyFields` has no values, the operation will be skipped. If it does need to run, the execution method has also been optimized.

HEIHUAa:
Optimized `Strum.hx`: `lastDrawCameras = cameras.copy();` under the very common scenario where `cameras.length == 1`.

MRK2009 (approved):
So pro 🔥

HEIHUAa:
Okay, fixed a previous issue with the code implementation.

Originally, the code directly restored the position within `isOnScreen`, causing incorrect positioning during rendering. I added two new external variables to cache this operation and restored it in `drawComplex`. Since `isOnScreen` was modified, I created a new `isOnScreenOriginal` method to replace the original public method.

Raltyro:
Is this okay to merge now?

HEIHUAa:
> Is this okay to merge now?

I think it's ready to be merged.

PR #952, opened 2026-04-06
Replace Nix shell configuration with a Flake

r6915ee:
This is a reiteration of #771 that replaces the original `shell.nix` file with a new [Nix Flake](https://nixos.wiki/wiki/flakes). The page linked describes the benefits of using Flakes.

The Flake is resistant to the deprecation of `mbedtls_2` by using an overlay. There are comments about this in the Flake itself.

A legacy `shell.nix` file is still included, but it's wired to use `flake-compat`.

In addition, the build inputs have also been altered to remove unnecessary packages and to adhere to the fairly recent removal of the `xorg` package set, where the packages inside of it were renamed.


r6915ee:
Currently, the Flake still has the same issue as the original pull request where it uses Haxe v4.3.6 instead of Haxe v4.3.7, primarily because Nixpkgs still hasn't updated its Haxe version. However, I've seen Codename being able to compile and run without issue, so I'm currently keeping it open as a non-draft.

r6915ee:
[This pull request](https://github.com/NixOS/nixpkgs/pull/508350) at Nixpkgs addresses the issue about the Haxe version. Currently, binaries are only cached for later versions of Darwin, but once I see that Haxe has been cached for most other platforms, I'll update the `flake.lock` file on my end.

r6915ee:
This pull request now properly uses Haxe v4.3.7. This also means that the `mbedtls_2` override workaround has been removed, because the current version of the Haxe package in Nixpkgs now depends on standard `mbedtls`.

r6915ee:
I'll be closing this pull request, since I no longer use Nix and thus I'm unable to test new changes. That being said, I'll try to at least keep the branch up if anyone wants to continue work on this.

PR #953, opened 2026-04-06
Add sum calculation functionality to dev CLI setup subcommand

r6915ee:
The dev CLI's setup subcommand has a new feature that allows it to calculate the sha256 hash of the current libraries file, storing it into an ignored `.sum` file. Without the new `--ignore-sum` flag, if the hash of the current file and the hash presented in the sum file are the same, then the setup process won't continue. The primary use case of this that I see is to avoid having to reinstall the libraries in the case of #952, where the script would always be run every time the development shell was entered.

From the commit message:

> The subcommand uses this feature to compare the sha256 sums of the previous and current
> building/libs.xml files in order to decide whether or not to initiate the setup
> process. It adds an `--ignore-sum` flag to immediately pass the check used for this.
> 
> This also clears up the help message for the subcommand in question, updates both
> `setup-` scripts to allow for variable arguments, and adds `--all` as an alias for
> `--reinstall`.


Raltyro:
.sum?

r6915ee:
> .sum?

Those files simply contain the last sha256 hash that was created from the `libs.xml` file and are purely local. The subcommand reads the last hash from this file and compares it to the respective hash of the current `libs.xml` file to check whether or not `libs.xml` was modified.

Raltyro:
Looks good, although --ignore-sum need to be implemented in workflows

PR #954, opened 2026-04-07
"Swap Event Sides" option for charter

Bitto1070:
Adds an option to debug settings that makes global events appear on the left.
I swear this was promised at some point but then never actually done. I can't seem to find the poll for it though so like. whatever man it's here
Again, I don't know polish or italian, and I don't think google translate is going to help this time.

Preview:
preview


Raltyro:
I don't know about this Bitto

Bitto1070:
what's wrong with it

PR #955, opened 2026-04-07
fix zoom event to actually set to the final zoom when complete

Bitto1070:


Raltyro:
Looks like something that should be handled in flixel side

PR #957, opened 2026-04-08
round stage editor position, scale, and angle values

Vechett:


HEIHUAa:
I think maybe directly modifying the drag to produce floating-point values might be better? And allow high precision to be manually set?

PR #958, opened 2026-04-09
Fix a minor issue in StrumLine

HEIHUAa:
Previously, using `str.ID` to read the state might have been inconsistent with the input from the code above:

```haxe
for (i in 0...membersLength) {
    __pressed[i] = members[i].__getPressed(this);
    __justPressed[i] = members[i].__getJustPressed(this);
    __justReleased[i] = members[i].__getJustReleased(this);
}
```

It has now been unified to use the `members` index. Additionally, the new approach may offer better performance.

### DIFF
--- a/source/funkin/game/Note.hx
+++ b/source/funkin/game/Note.hx
@@ -77,8 +77,9 @@ class Note extends FlxSprite
 
 	public var strumID(get, never):Int;
 	private function get_strumID() {
-		var id = noteData % strumLine.members.length;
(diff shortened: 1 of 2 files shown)


HEIHUAa:
Added an extra optimization: due to the frequent calls to `strumID`, the modulus operation for each call has been eliminated. Using an `if` condition is much faster than `%`.

Raltyro:
I'll check back on this to fix something about this

PR #959, opened 2026-04-10
Use default branch instead of use a specific version to avoid Node.js deprecations in the workflows

MRK2009:
I forgot to update those, because I checked in the main branch and it does affect on both of them due to Node.js version deprecation. even though they were disabled for now in the internal merge branch.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           haxe-version: 4.3.7
       # - name: Restore existing build cache for faster compilation
-      #   uses: actions/cache@v4.2.3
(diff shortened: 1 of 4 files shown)


MRK2009:
Wait let me test it out on the main branch first, before let this merged

MRK2009:
Yep, it does work.

MRK2009:
I'm still seeing Node.js version deprecations somewhere else

MRK2009:
I think I'm done.

MRK2009:
I guess I'm giving up, because it confuses me

Raltyro:
I might look at this more into it later

PR #963, opened 2026-04-16
Fix error caused by modifying Flags.MAX_SPLASHES

HEIHUAa:


source/funkin/game/SplashGroup.hx
```diff
@@ -92,8 +92,7 @@ class SplashGroup extends FlxTypedGroup<Splash> {
 	}
 
 	function pregenerateSplashes(splash:Splash) {
-		// make 7 additional splashes
-		for(i in 0...7) {
+		for (i in 0...Flags.MAX_SPLASHES - 1) {
```
Raltyro, line 96: Should be Math.min(Flags.MAX_SPLASHES, 7) i think? so it doesnt make alot if MAX_SPLASHES is 32 or above

Raltyro (changes requested):

HEIHUAa:
If we want to change it to Math.min(Flags.MAX_SPLASHES, 7) without crashing, we'll have to dig into flixel.group to find the real issue.

Raltyro:
Do you have the crash log? it would be helpful

HEIHUAa:
> Do you have the crash log? it would be helpful

Wait a moment.

HEIHUAa:
> Do you have the crash log? it would be helpful

2026-08-01-13-05-18-CodenameEngine
This is the error pop-up. The crash occurs when the first splash effect is triggered.

HEIHUAa:
At least judging from the code, I believe that `FlxTypedGroup.recycle()` fails to retrieve a splash, which leads to `__splash.strum = strum;` — however, the crash occurs at the point of assignment.

Raltyro:
can you confirm if this crash also happens in internal-merge?

HEIHUAa:
> can you confirm if this crash also happens in internal-merge?

Yes, internal-merge still has this issue.

1813780f-9625-4c9f-bcba-37b0576a2a22

HEIHUAa:
2026-08-01-17-26-47-CodenameEngine

I also tried the freshly built internal-merge, and the issue remains unchanged.

Raltyro:
For now i'm merging this

PR #964, opened 2026-04-16
EventManager-Optimization

HEIHUAa:
Given that this needs to be triggered multiple times per frame, I think it's necessary to change the original O(n) operation to O(1), along with platform-specific optimizations for CPP and adding an inline option.


Raltyro:
Great stuff, i'll test this if i get the time to compile CNE myself

PR #965, opened 2026-04-19
Fixed charter editors

HEIHUAa:
**Fixed an issue in charter editors where, after losing focus from the right side of the window, clicking directly on the center of the window would automatically modify the timeline.**


novathepurplecord:
YES MERGE THIS PLEASE 🙏

MRK2009 (approved):

PR #968, opened 2026-04-29
Mods are sorted alphabetically on all platforms.

HELLISH666-warframe:
Very intricate code added right here!!111!!one11!!!

### DIFF
--- a/source/funkin/menus/ModSwitchMenu.hx
+++ b/source/funkin/menus/ModSwitchMenu.hx
@@ -30,6 +30,7 @@ class ModSwitchMenu extends MusicBeatSubstate {
 		FlxTween.tween(bg, {alpha: 0.5}, 0.25, {ease: FlxEase.cubeOut});
 
 		mods = ModsFolder.getModsList();
+		CoolUtil.sortAlphabetically(ModsFolder.getModsList(),true);
 		mods.push(null);
 
 		alphabets = new FlxTypedGroup<Alphabet>();


source/funkin/menus/ModSwitchMenu.hx
```diff
@@ -30,6 +30,7 @@ class ModSwitchMenu extends MusicBeatSubstate {
 		FlxTween.tween(bg, {alpha: 0.5}, 0.25, {ease: FlxEase.cubeOut});
 
 		mods = ModsFolder.getModsList();
+		CoolUtil.sortAlphabetically(ModsFolder.getModsList(),true);
```
r6915ee, line 33: I would probably shorten this line and line 32 into one by doing:

```haxe
mods = CoolUtil.sortAlphabetically(ModsFolder.getModsList(), true);
```

Because `sortAlphabetically` returns the modified array, it's safe to directly use it as the value of `mods`.

r6915ee (commented):
The review comment attached below is primarily for if the pull request goes down this route.

My only major complaint is that this will sort the list _each time the mods menu is opened_. This might not affect small mods lists that much, but if the player has many mods installed, then this may be able to cause a noticeable slowdown.

What I would instead do is memoize the sorted list, alongside a hash of the concatenated mod directory names (preferably through an SHA-2 variant) and the last sorting mode used. Sorting and memoization would instead occur in `getModsList`. The last hash is compared against the current hash alongside the last sorting mode, and if any of them differ, then sorting would reoccur.

This could perhaps be furthered by storing these values together alongside with other variants, further improving performance. The main drawback would be increased memory usage, but memoization can perhaps be configured to use only a user-defined amount of entries.

MRK2009:
@HELLISH666-warframe are you sure hmm?

PR #971, opened 2026-05-01
Add mod list sorting system

r6915ee:
The mods list can now be sorted in-engine with a cache to come with it.

From the commit message:

> ModsFolder.getModsList now has an optional parameter to describe the various sorting
> options. Excluding this option produces rather legacy behavior and should be
> backwards-compatible.
> 
> Sorted mod lists get cached under a so-called "sort forge," which writes the key to use
> in the cache. If a cache hit occurs, then ModsFolder.getModsList will use the hit.
> 
> Currently, there is support for ascending/descending order and two modes: clean and
> alphabetical sorting. Clean mode provides legacy behavior using this new system, and
> alphabetical sorting uses CoolUtil to sort the list.
> 
> ModSwitchMenu has been updated to use this new system, but really, the only noticeable
> change should just be additional memory overhead.

The primary use case for this would be for extending the mods switch menu to support multiple sorting options, all while being performant in the long run, especially with larger mods lists. The caching in particular assists in this by avoiding the overuse of sorting, doing so only once.

The system should be able to add more sorting mechanisms easily, just by adding the mode to the main enumerator and adding support for the mode in `ModSortingController`. Of course, this does believe that the raw output of the sort is in ascending order, as descending order takes higher priority. These likely will need to come in succeeding pull requests, though, since (as of writing) I don't have any plans to add additional sorting modes in this pull request.

### DIFF
--- a/source/funkin/backend/assets/ModsFolder.hx
+++ b/source/funkin/backend/assets/ModsFolder.hx
@@ -120,11 +120,7 @@ class ModsFolder {
 
 			sortForge.add(Std.string(sortingOptions.descending ? 1 : 0));
 
-			final modeId:String = switch (sortingOptions.mode) {
-			    case ModSortingMode.CLEAN: "c";
-				case ModSortingMode.ALPHABETICAL: "a";
-			};
-			sortForge.add(modeId);
+			sortForge.add(sortingOptions.mode);
 			
 			final sortForgePure:String = sortForge.toString();
 			final sortForgeHashed:String = haxe.crypto.Sha256.encode(sortForgePure);
@@ -234,14 +230,14 @@ class ModSortingController {
 /**
  * The mods list can be sorted in all of the ways provided by this enum.
  */
-enum ModSortingMode {
+enum abstract ModSortingMode(String) {
     /**
      * Use the original list received from reading the directory. This may depend
      * on the current platform, but remains for legacy purposes.
      */
-    CLEAN;
+    var CLEAN;
     /**
      * The list should be in alphabetical order.
      */
-    ALPHABETICAL;
+    var ALPHABETICAL;
 }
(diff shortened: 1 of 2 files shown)


r6915ee:
This is currently set as a draft pull request, because there's one tiny change I want to make, and that is to move the mode ID function over to the controller. However, it's likely going to come tomorrow where I'm at, since it's getting far too late as I'm writing this.

r6915ee:
The mode ID is now assigned by using a string abstract enum, alongside a new delimiter for each mod in the sort forge.

Raltyro:
Is there any example usage of this?

r6915ee:
> Is there any example usage of this?

Sorting does get used in the [regular mod switch menu here](https://github.com/CodenameCrew/CodenameEngine/pull/971/changes#diff-dbbf6cc20336107d8aa7bda3b81dcfbd076a3a3ced555fd987507159c234c3bf) if that's what you're referring to, though it essentially functions the same way it did before the change, since there's no added way to change the sorting ingame at the moment:

```haxe
mods = ModsFolder.getModsList({
    descending: false,
    mode: CLEAN,
});
```

Raltyro:
If we're considering this, might aswell do it to addons too and add a way to change the sorting here

r6915ee:
> If we're considering this, might aswell do it to addons too

Hm, addons are currently just treated as libraries and `ModsFolder` doesn't care about them, so I'll probably make a pull request that allows `ModsFolder` to keep track of the list of addons (and _maybe_ move the sorting from this pull request to over there, as there's a few things I want to polish up) in conjunction to or after this one. I do have a couple of ideas on how to do that without breaking changes to the API though, primarily just adding the necessary fields and methods to both `MainState` and `ModsFolder` for this.

Raltyro:
Is this good to go?

r6915ee:
> Is this good to go?

There's still no option to switch what sorting to use in the mod switch menu, but other than that, it should be fairly good to go now and it works for me.

PR #974, opened 2026-05-02
Change the default value of "Multiplicative?" in the "Camera Zoom" event to be false.

ADA-Funni:


sillyburg:
i might be the only person ever to intentionally use multiplicative

Raltyro:
This piece of default value is fucking up over every mods ever created in CNE

PR #1000, opened 2026-06-07
Use environment variables to change the Haxe version on both `release` & `debug` easily and set hxcpp compile cache without the "~" symbol

MRK2009:
For adding `HAXE_VERSION` environment variable, y'all don't need to change the haxe version on both `release` and `debug` parts, instead y'all can  change the `HAXE_VERSION` to use a specific haxe version on both of them easily.

And for adding `HXCPP_COMPILE_CACHE` environment variable, this should make compilation a bit more faster(check https://github.com/openfl/lime/commit/8090bacdc6141c982a70dd8df381483829c04278 for more info about it)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -4,6 +4,10 @@ on:
   push:
   workflow_dispatch:
 
+env:
+  HAXE_VERSION: 4.3.7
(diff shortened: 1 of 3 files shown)


MRK2009:
LET'S GO I LEARNED HOW TO FORCE-PUSH!!!

MRK2009:
Thanks

PR #1015, opened 2026-06-26
Ratings event addition to the engine

arer-as:
https://github.com/user-attachments/assets/415c90e5-d178-4e72-bb41-ae5d0310e1d3

Implements a new event and two function calls for the ratings. RatingsShowEvent, onRatingsShown and onPostRatingsShown respectively.

### DIFF
--- a/assets/data/scripts/pixel.hx
+++ b/assets/data/scripts/pixel.hx
@@ -75,7 +75,7 @@ function onCountdown(event) {
 	};
 }
 
-function onPlayerHit(event:NoteHitEvent) {
+function onRatingsShown(event:RatingsShowEvent) {
 	if (!enablePixelUI) return;
 	event.ratingPrefix = "stages/school/ui/";
 	event.ratingScale = daPixelZoom * 0.7;
--- a/source/funkin/backend/scripting/events/gameplay/RatingsShowEvent.hx
+++ b/source/funkin/backend/scripting/events/gameplay/RatingsShowEvent.hx
@@ -1,0 +1,87 @@
+package funkin.backend.scripting.events.gameplay;
+
+import flixel.math.FlxPoint;
+
+final class RatingsShowEvent extends CancellableEvent 
+{
+	/**
+	 * Rating sprite (may be null)
+	*/
+	public var ratingSprite:Null<FlxSprite>;
+	/**
+	 * Number sprite (may be null)
+	*/
+	public var numberSprite:Null<FlxSprite>;
+	/**
+	 * Combo sprite (may be null)
+	*/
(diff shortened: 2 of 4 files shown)


r6915ee:
I've been thinking about adding something similar. That being said, are you sure that this won't break mods? Lines [1969 and 1971 in PlayState](https://github.com/CodenameCrew/CodenameEngine/pull/1015/changes#diff-6181a9bfd8531f9bffe728c38e6ae4545518b115306e76b2e7ff96d2e16a8b53R1969-R1971) are changed to remove certain parameters from the various `NoteHitEvent` instances, so that makes me think that may cause some complications.

arer-as:
> I've been thinking about adding something similar. That being said, are you sure that this won't break mods? Lines [1969 and 1971 in PlayState](https://github.com/CodenameCrew/CodenameEngine/pull/1015/changes#diff-6181a9bfd8531f9bffe728c38e6ae4545518b115306e76b2e7ff96d2e16a8b53R1969-R1971) are changed to remove certain parameters from the various `NoteHitEvent` instances, so that makes me think that may cause some complications.

This could break mods, but im not sure how i would make a failsafe for that

arer-as:
Although maybe only for displayRating/displayCombo? It's kind of a weird situation here. Maybe just add back the vars but they dont do anything? I dunno

arer-as:
> Although maybe only for displayRating/displayCombo? It's kind of a weird situation here. Maybe just add back the vars but they dont do anything? I dunno

realised i didnt upload the version of NoteHitEvent without those vars, not sure if i should now though

r6915ee:
> Although maybe only for displayRating/displayCombo? It's kind of a weird situation here. Maybe just add back the vars but they dont do anything? I dunno

I would probably just leave those two lines unchanged from the `main` branch, but keep everything else if they work. The original versions of the lines should hopefully still be compatible with the new additions from what I'm seeing here.

arer-as:
> > Although maybe only for displayRating/displayCombo? It's kind of a weird situation here. Maybe just add back the vars but they dont do anything? I dunno
> 
> I would probably just leave those two lines unchanged from the `main` branch, but keep everything else if they work. The original versions of the lines should hopefully still be compatible with the new additions from what I'm seeing here.

Just made some commits to support NoteHitEvent rating stuff. Hopefully its ok now

sillyburg:
pls merge this raltyro.. i need this... my ratings are kinda rtxless..

Raltyro:
whateven is rtx, sure

sillyburg:
> whateven is rtx, sure

idk that's the first thing i thought of

PR #1019, opened 2026-07-01
FunkinText zoomFactor fix

Bitto1070:
Fixes zoomFactor causing FunkinText to slowly drift when zoomFactor isn't 1 (hopefully).


Raltyro:
Not merging until proven it fully works? just concerned with the text "hopefully" here

Bitto1070:
> just concerned with the text "hopefully" here

it worked perfectly fine in testing but you never know when `niche use case` shows up and breaks everything

PR #1021, opened 2026-07-07
stage editor decimal rounding

clairedeluneee:
yall dont need this amount of precision dawg
also dont mind the newlines zed did that on its own 🥀


Raltyro:
Looks good enough, although list of things i wanted to do before merging this
- any 0 decimals using FlxMath.roundDecimal should just use Math.round instead
- change any decimals (1, 2) to just 3 instead

MRK2009:
> Looks good enough, although list of things i wanted to do before merging this
> 
> * any 0 decimals using FlxMath.roundDecimal should just use Math.round instead
> * change any decimals (1, 2) to just 3 instead

@clairedeluneee ☝️

HEIHUAa:
I think supporting high precision is correct — there's always the unexpected. And I know the real issue is the precision loss when dragging. I think directly fixing the xy values during drag is the more reasonable approach.

clairedeluneee:
someone forgot to read her notifications it seems

MRK2009:
XD

Raltyro:
Okay maybe if the issue should be fixed in the editor instead of saving, because i don't really wanna restrict precisions even if it's for some reason

Raltyro:
Closing this as it seems not the right thing todo to just restrict the values when saving, it should be when editing the characters, make it snap to not have any decimals

Issue #1024, opened 2026-07-15
Compiling help: HScript IMPORT_BLACKLIST Compiling error


dseclipse:
[x] Windows
[ ] Linux
[ ] Mac
Tried compiling Codename from the base source and got problems with Hscript (Haxe 4.3.7)

<img width="1269" height="111" alt="Image" src="https://github.com/user-attachments/assets/c022f66c-2b10-4604-9a1e-a8daeab6e802" />
All the libs are the base ones on the libs.xml 
I tried changing the Hscript libs(master and archive/codename-dev branches) but unsuccessful

r6915ee:
Oh, that's particularly interesting. [This change in `hscript-improved`](https://github.com/CodenameCrew/hscript-improved/commit/266bb67c9361767cbfff66280a49b4a7d3269fdc) is causing the issue.

Codename overrides the `Config` class, and it doesn't seem to have the `IMPORT_BLACKLIST` field that was introduced. I can perhaps get a fix for that in as soon as possible, I have some spare time on my hands available.

Jamextreme140:
god, I forgot about it. Hold on...

Jamextreme140:
It's fixed now :3

PR #1026, opened 2026-07-15
Add IMPORT_BLACKLIST to Config.hx 

Jamextreme140:
small change due to changes in hscript library

this fixes the issue #1024


SrtHero278:
well i dont see why not

PR #1031, opened 2026-07-22
Add angleFactor to FunkinSprite and FunkinText

BloomEld:
basically the same thing zoomFactor does but for angle probably useful


Raltyro:
I'm kinda not sure, does this works fully for both FunkinText and FunkinSprite?

BloomEld:
> I'm kinda not sure, does this works fully for both FunkinText and FunkinSprite?

i've used it for 2 mods so far, and it works fine. i haven't found any bugs with it.
https://github.com/user-attachments/assets/f013b6ac-2390-43ac-b10b-60c3136baa6d

PR #1033, opened 2026-07-25
Hardcoded scripts (wip)

TheZoroForce240:


r6915ee:
I'm quite curious, what would be the main use case of this? I can see this having some potential, but I'm somewhat concerned how this will work for different platforms.

TheZoroForce240:
It's mainly meant for large mods with lots of big scripts, where hscript just isn't fast enough to handle it.
You can either setup your mod to use a hybrid method where you use hscript to make/debug the script and then use a flag to auto generate source files when finished, or just do the hardcoded scripts manually, I've designed it to not be too different from using hscript and still supports things like public and static script vars.
For getting it working on other platforms, it should be as simple as forking the cne repo, edit project.xml to enable hardcoded scripts, drop the script files in the modSource folder and then run workflows. I've got wip documentation for it I'll pr to the wiki as well with how it should be used

A basic script looks something like this:
```haxe
@:registerScript("songs/bopeebo/scripts/script.hx")
class Script extends HardcodedScript<PlayState> {

	var a:Int = 0;
	
	@:publicVar var testPublic:Float = 50.0;  //@:publicVar generates a get/set func to directly access the publicVariables map
	
	@:publicVarRef var testPublicFromAnotherScript:Float; //same as @:publicVar but doesn't init the variable, mainly for helping the compiler know which type it is

	public function create() {
		trace("script loaded");
	}

	override public function update(elapsed:Float) {
		parent.dad.angle += testPublic*elapsed;
	}
	
	function onNoteHit(event:NoteHitEvent) {
		trace("note hit");
	}
}
```

Raltyro:
Might actually be really useful,
although i was kinda hoping you could like just use the codes from a ndll, but it wouldnt make sense now that i think about it

MRK2009:
It got conflicted

PR #1036, opened 2026-07-28
Added to Paths with meta support

drfiffi:
better pathing for meta data


Raltyro:
why, couldnt you also make it so it replaces like any usage to this

SrtHero278:
image
also fyi, -difficulty exists

PR #1038, opened 2026-08-01
(Scrapped) Stage class rewrite + Stage Editor Overhaul

Jamextreme140:
`Stage` class rewritten with many improvements along with an Overhauled Stage Editor that fixes many issues.

New Stage Features:

- Modularity: `Stage` now can be used outside of PlayState without any issues, i.e. for example, you can set a Stage in a custom state.
- Layering: Now you can set a group of sprite and easily organize them.
- Rendering: Just like `flixel-animate`, it allows you to set a shader, blend modes or `alpha` property to the entire stage (or layer) as a single sprite. Shoutouts for `MaybeMaru` for the rendering method :3.
- Stage Preloading: since `Stage` is now modular, it can be preloaded and switched mid-song without issues.

New Misc. Features:

- Change Character & Change Stage event.

PR by @Jamextreme140 & @ItsLJcool


MRK2009:
Cool

Jamextreme140:
I'm re-doing this...

PR #1049, opened 2026-08-05
VideoCutscene: Change the `video` type from `var` to `final`

MRK2009:
I forgot to include this in the other pr earlier, this is basically a small clean-up.

I tested the V-Slice modpack(which uses VideoCutscene) and it looks okay.

I checked in the hxvlc docs and it used to have `var` in it, but it is `final` recently.


MRK2009:
I swear

PR #1054, opened 2026-08-07
Add FunkinBackdrop

azmlpoh:
I hope FlxBackdrop can have zoomFactor to give mod creators more room for development.

I'm not sure if it's feasible. (


Raltyro:
dude atp we should just add anything as funkin, or move these zoomFactor/angleFactor to the base classes

sillyburg:
> dude atp we should just add anything as funkin, or move these zoomFactor/angleFactor to the base classes

i agree with the latter

Raltyro:
oh

MRK2009:
Are you serious?

PR #1056, opened 2026-08-09
Charter: Auto-generate strumlines if the JSON is missing

clairedeluneee:
2 am pull request. Testing highly encouraged. Maintainers are free to edit if needed.

Will be going to bed as soon as I hit the create pull request button. Goodnight.

### DIFF
--- a/source/funkin/editors/charter/Charter.hx
+++ b/source/funkin/editors/charter/Charter.hx
@@ -653,8 +653,8 @@ class Charter extends UIState {
 				vocalsSuffix: "",
 				keyCount: 4
 			};
-			createStrumLine(0, dadStrumLine, false, false);
-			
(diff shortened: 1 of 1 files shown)


clairedeluneee:
hi 7pm claire here with more mental clarity this time im actually gonna test this now if you dont mind

clairedeluneee:
ok it works but its crashing when i open the generated strumlines time to triage

clairedeluneee:
ok i think it works now

moxie-coder (approved):

clairedeluneee:
should have probably described what this does last week when i made the pr

basically, the charter does not check if the chart json is missing (either it was deleted, moved, or just never existed in the case where the song was just created) which leads to the game crashing if you do anything other than make a strumline
so what this does is that if it is actually missing the charter makes the two strumlines for you

PR #1062, opened 2026-08-12
Replace memoryPeak with actual process memory usage

HEIHUAa:
Replace the second memoryPeak with the actual memory usage of the process as reported by the operating system. I believe this will be significantly more useful than the previous peak memory value. This works on Windows, macOS, and Linux. On non‑CPP targets or platforms other than these three, only a single memory number will be displayed, with the second one omitted.

### DIFF
--- a/assets/languages/en/Options.xml
+++ b/assets/languages/en/Options.xml
@@ -206,6 +206,9 @@
 
 		<str id="songOffsetAffectEditors-name">Offset in Charter</str>
 		<str id="songOffsetAffectEditors-desc">If checked, this will enable the Song Offset option in the Chart Editor too.</str>
+
+		<str id="legacyMemoryCounter-name">Legacy Memory Counter</str>
+		<str id="legacyMemoryCounter-desc">If checked, the memory counter will show the Haxe GC memory usage and its peak (the original behavior) instead of the working set memory seen in Task Manager.</str>
 	</group>
 
 </language>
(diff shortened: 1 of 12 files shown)


r6915ee:
I think it may be useful to keep around an option to enable the peak memory value, as I've used it on occasion to determine memory usage issues in the case where a song might accidentally use too much memory and the current garbage-collected memory value changed too fast to get the actual value. Otherwise, I think this can also be useful.

HEIHUAa:
> I think it may be useful to keep around an option to enable the peak memory value, as I've used it on occasion to determine memory usage issues in the case where a song might accidentally use too much memory and the current garbage-collected memory value changed too fast to get the actual value. Otherwise, I think this can also be useful.

Okay, I think your suggestion is fair. I'll add a setting to the debug options later. Personally, I've never really paid attention to the peak memory value — most of the time, it just gets frozen after a single large spike and stays there for a long time, which makes it pretty much useless except for very specific development needs. In most cases, it doesn't show anything meaningful.

HEIHUAa:
I added a setting item in the DEBUG OPTIONS menu, but I'm not sure if the naming is appropriate. It is now ready for review and commit.

Raltyro:
i'm against with the linux implementation here? is there not a direct way to get the value?

r6915ee:
> i'm against with the linux implementation here? is there not a direct way to get the value?

The other methods for getting the `VmRSS` value I know of seem to require root privileges.

That being said, I do think that implementation in specific could be changed to only update the value every now and then to avoid polling that file, as the file comes from the kernel.

HEIHUAa:
I think obtaining the RSS is the more correct method, and there doesn't seem to be any better alternative. I've added a throttle that only polls the kernel file once every 0.5 seconds.

If there's a better way, you're welcome to suggest it.

PR #1069, opened 2026-08-15
Remove workaround for Away3D View which is no longer necessary

kitzsh:
The issue seems to have fixed itself somewhere during 1.1.0 development, interesting.
Also any width/height value below 0 will now make the view's size match the screen, rather than just -1.


kitzsh:
Meant for internal-merge, not main, whoops

PR #1072, opened 2026-08-16
add LimeAssets to Script imports

THEkatinamicrowave:
This sucks and I don't like doing this this legitimately took me hours to figure out
image

So I'm pleading that you add `LimeAssets` to the thing

Also please make `Assets.getBitmapData()` returning an `OptimizedBitmapData` be optional and like toggleable off in the function somehow or expose `image` somehow

### DIFF
--- a/source/funkin/backend/scripting/Script.hx
+++ b/source/funkin/backend/scripting/Script.hx
@@ -51,6 +51,7 @@ class Script extends FlxBasic implements IFlxDestroyable {
 			// OpenFL & Lime related stuff
 			"BlendMode"			=> CoolUtil.getMacroAbstractClass("openfl.display.BlendMode"),
 			"Assets"			=> openfl.utils.Assets,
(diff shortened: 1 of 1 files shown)


betpowo:
`Assets.getBitmapData` arguments are `(file, useCache, useGPU)` already but sometimes itll not work and I Dont Know Why   . a method that Consistently works in all versions of cne is setting `Main.forceGPUOnlyBitmapsOff = true;` before calling getBitmapData and then setting it back to normal. i also wish this wasnt the case tho

SrtHero278:
you know codename has a `import lime.utils.Assets as LimeAssets` right?
also yeah Assets.getBitmapData has more parameteters.
`public static function getBitmapData(id:String, useCache:Bool = true, ?allowCompressedTextures:Null<Bool>, ?allowHardwareTextures:Null<Bool>):BitmapData`

PR #1078, opened 2026-08-18
Fix issue where fragmentPrefix and vertexPrefix in FunkinShader do not take effect

HEIHUAa:
Fixed an issue where `fragmentPrefix` and `vertexPrefix` did not take effect in 1.1.0.

### DIFF
--- a/source/funkin/backend/shaders/FunkinShader.hx
+++ b/source/funkin/backend/shaders/FunkinShader.hx
@@ -148,9 +148,40 @@ class FunkinShader extends FlxRuntimeShader implements IHScriptCustomBehaviour {
 	}
 
 	#if REGION /* Deprecated */
-	public var shaderPrefix:String = "";
-	public var fragmentPrefix:String = "";
-	public var vertexPrefix:String = "";
+	public var shaderPrefix(get, set):String;
+	public var fragmentPrefix(get, set):String;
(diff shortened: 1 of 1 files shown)


Raltyro:
Where is these prefixes being applied to? Its on purpose that this is being deprecated, although if anyone wants to make a pull request on trying to getting it back, you can

HEIHUAa:
> Where is these prefixes being applied to? Its on purpose that this is being deprecated, although if anyone wants to make a pull request on trying to getting it back, you can

Well, I don't quite understand. This seems quite useful – why is it being deprecated? I would use an externally supplied `#define` for this, because it allows more optimization compared to using a uniform-driven approach. For example, if there's a `for` loop that uses this value, it may even enable loop unrolling. If it's removed, we would have to manually write everything from scratch for this functionality.

Raltyro:
what was i even saying here

Raltyro:
Oh right, i think it wont be much optimized as using #include for Shaders in internal-merge, as it now properly caches shaders with the path it loaded from, the thing with shader ive changed is that ive made it automatically uncacheable if any shader code is modified after loading, but again if anyone are willing to make it work again, i dont mind

You can check that in cne-flixel-addoms FlxRuntimeShader

PR #1080, opened 2026-08-20
Add `excludeList` to FunkinParentDisabler

THEkatinamicrowave:
something that should probably be there. Could also make `ParentDisableable` typedef that just encompasses `<FlxTween, FlxCamera, FlxTimer, FlxSound>` but I think `Dynamic` is fine for now. Could break if somebody ever used `FunkinParentDisabler`'s first two constructor args but I think it's ok because I've never seen it used in mods (personally). If it does break too much stuff, can simply move `excludeList` to the end of the constructor args


source/funkin/menus/PauseSubState.hx
```diff
@@ -40,8 +42,10 @@ class PauseSubState extends MusicBeatSubstate
 
 	private var __cancelDefault:Bool = false;
 
-	public function new(?items:Array<String>, ?selectCall:NameEvent->Void) {
+	public function new(?excludeList:Array<ParentDisableable>, ?items:Array<String>, ?selectCall:NameEvent->Void) {
```
Raltyro: excludeList should be the last argument of constructor, having it as the first could break mods using the previous arguments

Raltyro (changes requested):

source/funkin/menus/PauseSubState.hx
MRK2009: > excludeList should be the last argument of constructor, having it as the first could break mods using the previous arguments

@THEkatinamicrowave ☝️

THEkatinamicrowave: bet

source/funkin/backend/utils/FunkinParentDisabler.hx
```diff
@@ -18,28 +18,31 @@ class FunkinParentDisabler extends FlxBasic {
 	var __cameras:Array<FlxCamera>;
 	var __timers:Array<FlxTimer>;
 	var __sounds:Array<FlxSound>;
+    var __excludeList:Array<ParentDisableable>;
 	var __replaceUponDestroy:Bool;
 	var __restoreUponDestroy:Bool;
-	public function new(replaceUponDestroy:Bool = false, restoreUponDestroy:Bool = true) {
+
+	public function new(excludeList:Array<ParentDisableable> = [], replaceUponDestroy:Bool = false, restoreUponDestroy:Bool = true) {
```
MRK2009: I think this includes for FunkinParentDisabler too, if I'm not mistaken.(Also sorry for the double notification, I pressed the wrong button.)

THEkatinamicrowave: What includes `FunkinParentDisabler`? `ParentDisableable`?

MRK2009: Yes.

THEkatinamicrowave: I'm unsure what use case could rationalize putting `FunkinParentDisabler` in `ParentDisableable`, but if you can find one I can probably do that. For the record, `ParentDisableable` is a wrapper for `FlxTween`, `FlxTimer`, `FlxCamera`, and `FlxSound`, since I don't want to make the compiler deal with `Dynamic`.

THEkatinamicrowave:
holy I think I finally fixed every error it's actually building

PR #1081, opened 2026-08-20
DEV_RELOAD clear Cache

HEIHUAa:
This PR is intended to address the issue in the internal-merge branch where most things are cached, preventing timely refreshes. Therefore, I think that when the DEV_RELOAD key refreshes the State, it should also be set to clear almost all caches. I believe this is a good solution.


Raltyro:
I don't think this is a good implementation, can causes crashes with multiple atlases, and i think it's intended normal reload to restart one state to not clear the cache except for switching mod

PR #1083, opened 2026-08-22
Add CLEAR_ASSET_CACHE_ON_STATE_SWITCH

HEIHUAa:
Since directly modifying the State reset might not be appropriate last time, I've changed it to add a new parameter CLEAR_ASSET_CACHE_ON_STATE_SWITCH in Flags to control whether to clear all caches when switching States. During mod development, it is necessary to update assets in real time.

### DIFF
--- a/source/funkin/backend/system/modules/FunkinCache.hx
+++ b/source/funkin/backend/system/modules/FunkinCache.hx
@@ -54,10 +54,9 @@ class FunkinCache extends AssetCache {
 	**/
 	public static function clearAllCaches() {
 		@:privateAccess
-		for (key in [for (k in FlxG.bitmap._cache.keys()) k]) {
-			var graphic = FlxG.bitmap._cache.get(key);
+		for (graphic in [for (g in FlxG.bitmap._cache) g]) {
 			if (graphic != null && graphic.assetsKey != null)
-				FlxG.bitmap.removeByKey(key);
+				FlxG.bitmap.removeByKey(graphic.key);
 		}
 		if (instance != null)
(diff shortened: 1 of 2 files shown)


source/funkin/backend/system/modules/FunkinCache.hx
```diff
@@ -42,6 +42,35 @@ class FunkinCache extends AssetCache {
...
+	/**
+		Clears every asset cache (textures, sounds, fonts, shader programs) so everything is re-fetched from disk, as if the game just started.
+	**/
+	public static function clearAllCaches() {
+		@:privateAccess
+		for (key in [for (k in FlxG.bitmap._cache.keys()) k]) {
+			var graphic = FlxG.bitmap._cache.get(key);
+			if (graphic != null && graphic.assetsKey != null)
+				FlxG.bitmap.removeByKey(key);
+		}
```
r6915ee: Unless this doesn't work as intended, wouldn't it be simpler and slightly more optimal to iterate over the actual `FlxGraphic` objects instead and fetch the key from there? The `FlxGraphic.key` field should already point to the actual cache key.

HEIHUAa: That makes sense. When I wrote that code, I wasn't actually focusing on that — instead, I was focused on figuring out how to clear the shader cache with a single button.

HEIHUAa: I made the changes according to your suggestion, and it should be fine now.

Raltyro:
not a fan...

PR #1092, opened 2026-09-04
"onBopZoom" Event

sillyburg:
Adds an "onBopZoom" callback which allows you to cancel and modify certain parameters when the zoom bop thing happens.

Also made it a separate function for events since I found it inconvenient.

https://github.com/user-attachments/assets/3e91c9c5-0eab-4cb4-95bd-3c2f936d187a


sillyburg:
i made this cuz i hate camera bop event.

PR #1098, opened 2026-09-11
fix thorns scripted cutscene softlock

joaquinLOL123:
Playstate.instance.SONG doesn't work anymore for some reason so it just locks up when loading the dialogue uh oh

### DIFF
--- a/assets/songs/thorns/cutscene.hx
+++ b/assets/songs/thorns/cutscene.hx
@@ -53,7 +53,7 @@ function create() {
 			else {
 				timers.remove(swagTimer);
 				camHUD.visible = true;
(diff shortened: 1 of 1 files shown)


betpowo:
it doesnt work anymore cus now its a static variable  . so yes PlayState.SONG would be the way to access it

Jamextreme140:
That thing worked because of a weird behavior performed in HScript that allows you to access static fields from an instance (workaround made for accessing static variables without referencing the class first) that has been disabled (and soon removed).

joaquinLOL123:
> That thing worked because of a weird behavior performed in HScript that allows you to access static fields from an instance (workaround made for vanilla custom classes implementation) that has been disabled (and soon removed).

I see, thanks for the insight, i was pretty confused at first why something like that would stop working all of a sudden.

PR #1105, opened 2026-09-13
Charter: Automatically rename Voices audio file if it's lowercase

doggywatty:
I was working on a mod of mine when i tried to edit one of the characters in the Chart Editor.
It crashes.
..and turns out it was an audio filename lowercase issue, so i made this PR.


Raltyro:
Not going to merge this because there is already a fixer for any paths that have different lowercases/uppercases for linux and mac, unless that doesnt even work somehow??

doggywatty:
Yeah, it doesn't work for me since im on Windows

Raltyro:
Your windows weird, it supposed to still able to read it regardless of what character cases is

doggywatty:
maybe because it's an optimized and modified version of windows, but it's whatever though

moxie-coder:
if you're using an modified Windows (mainly an "optimized" one) then that's not something we can exactly do about.

Something about that might've removed case insensitive paths but idk

Jamextreme140:
What "modified" windows are you using? Because Windows file system isn't case sensitive

doggywatty:
> What "modified" windows are you using? Because Windows file system isn't case sensitive

[FastOS 11 Pro 23H2](https://www.projectfastos.top/2025/03/fastos-11-23h2.html?m=1).

PR #1113, opened 2026-09-16
Fix console

HEIHUAa:
Fixed an issue where enabling console-related windows would cause a popup crash if a script called `__script__.destroy();`.


HEIHUAa:
Added some extra things; I wrote them a while ago and forgot to post them.

Optimized the code.
Cached uid/label.
Improved object display performance, and no synchronization is performed when an object is not expanded.
Added cache pre-splitting for log output, rebuilding only when the content changes.
Only draw visible lines.
Removed the ring buffer, replaced with push/shift to discard the oldest logs.
Completed a TODO to always follow the latest log.
Added Copy Log to the window content to solve the issue where the current version cannot copy.
Added a right-click menu to the window, allowing selection of Copy Everything / Copy Visible / Copy This Line / Clear (multilingual support not yet added).

HEIHUAa:
Optimized the console performance overhead that still occurs when the console is not visible (this issue significantly affected the maximum frame rate under unlimited frame rate).

PR #1118, opened 2026-09-17
mod and addon custom controls can now co-exist

### LINKED ISSUE #1116: Enhancement: IDK what title to give
Lets say a mod having a keybind for debugging called `Instant restart`, and an addon having a keybind called `Speed up gameplay`.

Those two effect the controls state, both of them gets added to the state, instead of just one.

_Basically just preventing this from happening:_
Image

_Don't mind the bg_
The file below is a from a mod, the one on the state is an addon.

clairedeluneee:
should fix #1116 in the case of both an addon and the currently active mod having a `controls.xml` of which the addon takes priority


clairedeluneee:
image
and yes you can in fact have multiple addons with custom controls

PR #1119, opened 2026-09-18
fix the caching problem (hopefully)

HeroEyad:
i just made a new `clearAll `function that wipes out everything when called

removes flixel bitmap entries that are no longer in use
reset the cache collections after clearing
reset `Paths.assetsTree'`s asset path cache alongside the main cache.


Raltyro (changes requested):
it still destroys the lime images even if the flixel usage is not 0 (although gpu cache doesnt store lime images anyway)

why is it in TranslationUtil.resetLanguage? and theorically alternatively if its in MainState, it is already doing that? unless im missing something here

HeroEyad:
> it still destroys the lime images even if the flixel usage is not 0 (although gpu cache doesnt store lime images anyway)
> 
> why is it in TranslationUtil.resetLanguage? and theorically alternatively if its in MainState, it is already doing that? unless im missing something here

ok ur right about the destroying part

on the placement, MainState already resets flixel's bitmap cache and the asset path cache before it calls setLanguage
so having it there is redundant on startup and mod switch. but switching language from the options menu never goes through MainState, it just calls setLanguage and reloadStrings, so FunkinCache never gets cleared there.


so would it be better to move the clear into LanguageRadio.select, right after setLanguage? that way it only runs when the user changes language or... idfk do you prefer something else like making a language change reload through MainState?

